PR #19, opened 2017-11-17
Fix #18: Drop elasticsearch dependency

### LINKED ISSUE #18: Upgrade Elasticsearch Dependency
The elasticsearch project has made many significant changes from version 1.7.0 to version 6.0.0, many of which were not backwards compatible and require potentially significant work to update rdf4j-sail-elasticsearch to work with a modern version of elasticsearch. Furthermore, the RDF4J SDK has not included rdf4j-sail-elasticsearch in over two years (not since Sesame 2.8.2) when it was last updated.

The last significant update to rdf4j-sail-elasticsearch was in Sesame version 2.8. Since then only the RDF4J APIs have been updated to version 2.2.3 and that version can be found here:
https://github.com/eclipse/rdf4j-storage/tree/b1cc29b7f8768b6551d8f77ee1991f37277383d9/elasticsearch

With effort underway to update Lucene version of RDF4J and to update RDF4J to jdk9, keeping rdf4j-sail-elasticsearch has become too expensive to justify its continued inclusion as-is and needs to be upgraded.

jamesrdf:
Signed-off-by: James Leigh <james.leigh@ontotext.com>

This PR addresses GitHub issue: #18 #7 eclipse/rdf4j#919.

* Drops elasticsearch as a dependency
* Makes it easier to upgrade lucene
* Makes it easier to upgrade Java


jamesrdf:
Yes, lucene and Solr are planned for 2.3.0. The trouble with keeping it (as is) is that it doesn't compile under jdk9.

jeenbroekstra:
Ah I see. In that case there's little else we can do I guess. You asked around, no one stepped up, so just drop it, and if someone squeeks ask them to help fix it :)

PR #24, opened 2017-12-12
Elasticsearch upgrade to 2.4.6.

pulquero:
Issue #18


pulquero:
I've just discovered that embedded nodes are no longer supported, i.e. can no longer start a node within java. This means it is going to be really hard to preserve any sense of backwards compatibility. Essentially you now have to start up your cluster yourself and the Sail will connect to it as a client.

pulquero:
The branch works fine, but there seems to be some problem with the merged result or travis, jar conflict or something being detected by the elasticsearch test classes.

pulquero:
That worked!

Issue #36, opened 2018-01-04
Make all sails faster by using hashing instead of trees


hmottestad:
SailSourceBranch uses a TreeModelFactory to provide a model to, presuamably store a transaction while it is on going.

I have been benchmarking a bit and found that switching to the LinkedHashModelFactory makes inserts in transactions 30% faster for the MemoryStore.

PR #37, opened 2018-01-04
Feature/36 linked hash model

hmottestad:
This PR addresses GitHub issue: #36 .

* Switch to using the LinkedHashModel for making inserts faster

I have benchmarked this change as part of the new SHACL sail, but I don't have any benchmarks on this branch.

### DIFF
--- a/sail-base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
+++ b/sail-base/src/main/java/org/eclipse/rdf4j/sail/base/SailSourceBranch.java
@@ -22,6 +22,7 @@ import org.eclipse.rdf4j.model.ModelFactory;
 import org.eclipse.rdf4j.model.Resource;
 import org.eclipse.rdf4j.model.Statement;
 import org.eclipse.rdf4j.model.Value;
+import org.eclipse.rdf4j.model.impl.LinkedHashModelFactory;
 import org.eclipse.rdf4j.model.impl.TreeModelFactory;
 import org.eclipse.rdf4j.query.algebra.StatementPattern;
 import org.eclipse.rdf4j.query.algebra.Var;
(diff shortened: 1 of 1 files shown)


hmottestad:
@jeenbroekstra @jamesrdf I was bechmarking the SHACL sail and found out that changing the SailSourceBranch to use a LinkedHashModelFactory makes it considerably faster (30%). Any reason why this is a bad idea?

jeenbroekstra:
LinkedHashModel and TreeModel are functionally identical, it's just that they cater for different access patterns. The first is a hash table, the second is a (balanced) binary tree. Average-case access for single search terms is O(1) for the hash, O(log(n)) for the tree. Worst-case access for hash tables is O(n) however, which can happen especially with searches on multiple terms (since LinkedHashModel does a sequential scan for the second term). 

So, if the Sail accessing the model typically does triple lookups with a single term,  a LinkedHashModel is clearly the superior choice. However if it's multiple-term lookups, it becomes a function of how your data is actually distributed (that is, if neither term significantly narrows the search space, the price of the sequential scan for the second term may be very high and therefore perform worse than the TreeModel). 

Clearly there's no single best answer for all cases. To be cautious you could override the implementation specifically for SHACL, since that seems to have a specific access pattern that suits the hashmodel better. I'm not sure that carries over the general case for SailSourceBranch though.

hmottestad:
@jeenbroekstra thank you for replying, I know you aren't able to be as active anymore so I really appreciate your input.

SailSourceBranch is used for handling transactions right? For storing added or removed statements on the transaction before they get committed.

So using a TreeModel would be good if you are causing queries on your added or removed statements. For instance when having multiple update queries. 

If I created a couple of benchmarks:

 - pure insert workload
 - pure remove workload
 - pure select workload
 - pure update workload
 - mixed workload (per transaction): select followed by insert
 - mixed workload (per transaction): insert followed by select (50% transactional data 50% base sail data)
 - mixed workload (per transaction): multiple update queries

Would also be good to know where this breaks down. Trees has a high cost for inserts (log n), lookups (log n) and iterations have a lot of cache misses. So there is likely an max N where the hashset is faster than the tree, and anything above N is only faster for pure insert/remoe workloads.

Issue #43, opened 2018-02-16
MemoryOverflowModel should use max heap size instead of currrently allocated RAM


jeenbroekstra:
The `MemoryOverflowModel` currently uses `Runtime.totalMemory()` to estimate memory consumption for determining whether to sync to disk. However, this method only returns the total _currently allocated_ memory, which is not necessarily the maximum amount of memory available for the JVM. This results in the model syncing to disk earlier than required, which has a large performance penalty. As part of a fix we should instead use `Runtime.maxMemory` which returns the maximum heap size, and do our treshold estimates from there.

hmottestad:
@jeenbroekstra I ran into this issue now on my #44, where the number of transactions created by the MemoryOverflowModel was so large that reducing the amount of bytes written to the transaction status file impacted the performance dramatically.

I have, however, found something you should look into. When I add System.gc() before the memory checks in MemoryOverflowModel, org.eclipse.rdf4j.sail.SailConcurrencyTest#testConcurrentAddLargeTxn goes from 84 seconds to 22 seconds. Nothing else changed!

```
	private synchronized void checkMemoryOverflow() {
		if (disk == null) {
			int size = size();
			if (size >= LARGE_BLOCK && size % LARGE_BLOCK == 0) {
				System.gc(); // Hint to JVM to do GC before we measure memory pressure
				long totalMemory = RUNTIME.totalMemory();
				long freeMemory = RUNTIME.freeMemory();
				long used = totalMemory - freeMemory;
				if (baseline > 0) {
					long blockSize = used - baseline;
					if (blockSize > maxBlockSize) {
						maxBlockSize = blockSize;
					}
					if (freeMemory < size / LARGE_BLOCK * maxBlockSize) {
						// may not be enough free memory for another
						overflowToDisk();
					}
				}
				baseline = used;
			}
		}
	}

```

hmottestad:
Btw. After some testing, this seems to be the correct way of measuring memory pressure in java:

```
long used = RUNTIME.totalMemory() - RUNTIME.freeMemory();
long freeMemory = RUNTIME.maxMemory()-used;
```

hmottestad:
Seems that the latter is the better fix.

jeenbroekstra:
@hmottestad that a call to `System.gc()` should make such a difference is a bit shocking. I've just put up a PR that doesn't use that trick, but does use Runtime.maxMemory. Can you verify if the performance improves as much with this change in place as well?

[...]

PR #45, opened 2018-02-17
GH-43 use total heap size and base on max block size

jeenbroekstra:
This PR addresses GitHub issue: #43 .

* use maxMemory to estimate total memory available for JVM
* sync to disk when when either not enough memory remains for another block or if less than 10% of the heap is still free (to avoid possible GC overhead errors)

@hmottestad  this is very close to what you suggested as well I think. I'm not keen on putting explicit calls to GC in the code but from local tests I ran this seems to significantly improve performance as well. Have a look, see what you think. 

PS for my next PR, I'll make sure I use the correct branch naming conventions again :)


hmottestad:
repository-sail/src/main/java/org/eclipse/rdf4j/repository/sail/helpers/SPARQLUpdateDataBlockParser.java:[99,27] error: incompatible types: Resource cannot be converted to BNode

hmottestad:
Merged develop into your branch, now it compiles. Test is down to 21 seconds for me (down from 84), and I checked the numbers in debug. Everything seems great. You should merge once Travis OKs it.

jeenbroekstra:
That test has been flaky and giving us intermittent grief for ages. I still haven't found the time to nail the issue down. But your restart seems to have magically fixed the issue (for now at least) so I'll go ahead and merge.

Issue #57, opened 2018-03-25
MemoryStore uses 30% of its time checking if statement exists, even if it doesn't need to notify anything


hmottestad:
`add(Resource subj, IRI pred, Value obj, SailDataset dataset, SailSink sink, Resource... contexts)` in `SailSourceConnection` checks if statements exists already, and if it does it notifies listening connections. It also returns `true` if statements were added.

From a quick check internally in this project, the return value is never used.

Dropping the return value and first checking if there are any listening connections should make inserts faster if there are no listening connections.

hmottestad:
Using Java 8.

## Before
```
Benchmark                        (isolationLevel)  Mode  Cnt   Score   Error  Units
MemoryBenchmark.simpleFoafModel              NONE  avgt   10  11.847 ± 0.352  ms/op
MemoryBenchmark.simpleFoafModel  READ_UNCOMMITTED  avgt   10  14.509 ± 0.613  ms/op
MemoryBenchmark.simpleFoafModel    READ_COMMITTED  avgt   10  14.555 ± 0.834  ms/op
MemoryBenchmark.simpleFoafModel     SNAPSHOT_READ  avgt   10  14.281 ± 0.683  ms/op
MemoryBenchmark.simpleFoafModel          SNAPSHOT  avgt   10  14.694 ± 0.735  ms/op
MemoryBenchmark.simpleFoafModel      SERIALIZABLE  avgt   10  27.086 ± 0.982  ms/op
```

## After

```
Benchmark                        (isolationLevel)  Mode  Cnt   Score   Error  Units
MemoryBenchmark.simpleFoafModel              NONE  avgt   10   5.864 ± 0.217  ms/op
MemoryBenchmark.simpleFoafModel  READ_UNCOMMITTED  avgt   10  11.122 ± 0.550  ms/op
MemoryBenchmark.simpleFoafModel    READ_COMMITTED  avgt   10  11.022 ± 0.510  ms/op
MemoryBenchmark.simpleFoafModel     SNAPSHOT_READ  avgt   10  11.222 ± 0.480  ms/op
MemoryBenchmark.simpleFoafModel          SNAPSHOT  avgt   10  11.167 ± 0.521  ms/op
MemoryBenchmark.simpleFoafModel      SERIALIZABLE  avgt   10  11.069 ± 0.460  ms/op
```

hmottestad:
[...]

PR #58, opened 2018-03-25
Issues/#57 check for listening connections first

hmottestad:
This PR addresses GitHub issue: #57 .

* Created a simple benchmark for the memory store
* Added check for if there are any connection listeners before calling hasStatement(....) when adding statements in the SailSourceConnection class
* Makes inserts for isolation level "NONE" 2x faster, "SERIALIZABLE" 2.5x faster and all others 1.3x faster

### DIFF
--- a/memory/pom.xml
+++ b/memory/pom.xml
@@ -67,6 +67,21 @@
 			<artifactId>junit</artifactId>
 			<scope>test</scope>
 		</dependency>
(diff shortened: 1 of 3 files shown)


hmottestad:
@jamesrdf what do you think, will this break something I'm not aware of elsewhere?

Issue #53, opened 2018-03-16
SHACL - support sh:datatype


hmottestad:
Reference to standard: https://www.w3.org/TR/shacl/#DatatypeConstraintComponent

Add support and tests.

hmottestad:
<img width="522" alt="screen shot 2018-03-16 at 17 42 18" src="https://user-images.githubusercontent.com/797185/37533170-69b5088e-2941-11e8-97a3-a6a501759eb7.png">

Initial validation plan

hmottestad:
done

PR #68, opened 2018-04-05
Features/53 shacl sh datatype

hmottestad:
This PR addresses GitHub issue: #53  .

* support for sh:datatype
* some cleanup of nodes so that creating a plan doesn't read any data
* more nodes for doing joins
* access to two connections, one which is the current connection, and one which is a connection with a transaction started right before the current one, so the validation plan can query for data that was in the sail from before.


hmottestad:
I've forgotten to sign my commits, need to fix that.

shacl/pom.xml
```diff
@@ -32,6 +32,12 @@
             <artifactId>rdf4j-sail-memory</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>rdf4j-queryparser-sparql</artifactId>
+            <version>2.3-SNAPSHOT</version>
+        </dependency>
```
hmottestad: Need this to force use of the fixed query parser.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/MaxCountPropertyShape.java
```diff
@@ -63,7 +63,7 @@ public PlanNode getPlan(ShaclSailConnection shaclSailConnection, Shape shape) {
 			planAddedStatements1 = new LoggingNode(((TargetClass) shape).getTypeFilterPlan(shaclSailConnection, planAddedStatements1));
 		}
 
-		PlanNode mergeNode = new LoggingNode(new MergeNode(planAddedStatements, planAddedStatements1));
+		PlanNode mergeNode = new LoggingNode(new UnionNode(planAddedStatements, planAddedStatements1));
```
hmottestad: MergeNode has been renamed to UnionNode, since the union term is used in spraql. Mind you, it’s a distinct union operation.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/MinCountPropertyShape.java
```diff
@@ -86,11 +86,13 @@ public PlanNode getPlan(ShaclSailConnection shaclSailConnection, Shape shape) {
 			//topNode = new LoggingNode(new BulkedExternalLeftOuterJoin(unique, shaclSailConnection.addedStatements, path.getQuery()));
 
 		} else {
-			String query =shape.getQuery();
 
+			PlanNode planAddedForShape = new LoggingNode(shape.getPlanAddedStatements(shaclSailConnection, shape));
+
+			PlanNode select = new LoggingNode(new Select(shaclSailConnection.addedStatements, path.getQuery()));
+
+			topNode = new LoggingNode(new LeftOuterJoin(planAddedForShape, select));
```
hmottestad: Explicit left outer join instead of trying to build a sparql query. Building a sparql query with optional will likely be faster, but this is a task for optimisation stage (which needs to be implemented in the future).

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/DatatypePropertyShape.java
```diff
@@ -0,0 +1,88 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Eclipse RDF4J contributors.
```
jeenbroekstra: This is really minor but shouldn't this say 2018?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -49,8 +51,9 @@
 	private HashSet<Statement> addedStatementsSet = new HashSet<>();
 	private HashSet<Statement> removedStatementsSet = new HashSet<>();
 
-	ShaclSailConnection(ShaclSail sail, NotifyingSailConnection connection) {
+	ShaclSailConnection(ShaclSail sail, NotifyingSailConnection connection, NotifyingSailConnection separateConnection) {
 		super(connection);
+		this.separateConnection = separateConnection;
```
jeenbroekstra: Can we rename this to `previousStateConnection` so it's more clear from the code what the purpose of this secondary connection is? Also: why do you even need to do this at this point? You have access to the sail object, can't you simply open a separate connection if and when your validation needs it? As long as your current connection hasn't committed, it should reflect the correct state no matter when you open the connection.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -38,8 +38,10 @@
 
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 
+	public NotifyingSailConnection separateConnection;
```
jeenbroekstra: Can we change this private with a separate protected (or public) accessor method?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -101,23 +110,30 @@ private SailRepository getNewMemorySail() {
...
-			boolean valid = validate();
-			if (!valid) {
-				rollback();
-				throw new SailException("Failed SHACL validation");
-			} else {
-				super.commit();
+		synchronized (sail) {
+			try {
+				boolean valid = validate();
+				separateConnection.commit();
```
jeenbroekstra: Perhaps I misunderstand the purpose of the separate connection - I thought it was just intended to read the state of the store before the current transaction? If so - why do you need to commit?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
hmottestad: yes, this is just me being lazy.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
hmottestad: Don't need to commit no. However, I thought I either had to commit or rollback once I start a transaction. The next time the current connection starts a new transaction, the separateConnection needs to start a new transaction too. In essence the separateConnection is a way of seeing the data as it was at the beginning of your current transaction, before you started modifying anything.

hmottestad:
@jeenbroekstra cleaned up the code you commented on. Had to rebase all commits to add sign-off.

About the .commit().

What is the best way that I can start a transaction on my `previousStateConnection` every time I start a transaction on the current connection? And how should I close the transaction on `previousStateConnection`, commit() or rollback()?

hmottestad:
@jeenbroekstra could you take a look at my last comment?

PR #70, opened 2018-05-01
eclipse/rdf4j#1113 shacl api refactor 

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#1113 .

* Removed separate repo for rules/shapes from Sail constructor
* Introduced named graph for storage/retrieval of shapes, gets ignored by default queries. 

Work In Progress, putting this up for early review: it's not ready for merge yet. In particular I want to take a closer look at the named graph approach and ensure that for SPARQL queries as well it accurately uses correct contexts.  Feedback welcome.


jeenbroekstra:
Ah right, I was wondering why there were so few tests - didn't realize there was a separate compliance and testsuite module. Will fix.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/DatatypePropertyShape.java
```diff
@@ -34,49 +35,52 @@
 
 	private final Resource datatype;
 
-	DatatypePropertyShape(Resource id, SailRepositoryConnection connection, Shape shape) {
+	DatatypePropertyShape(Resource id, ShaclSailConnection connection, Shape shape) {
```
hmottestad: Actually intended to use SailRepositoryConnection as much as possible, since that allows us to use the previousStateConnection interchangeably. Only where it was needed, was the ShaclSailConnection used.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
```diff
@@ -61,50 +62,46 @@ public boolean requiresEvaluation(Repository addedStatements, Repository removed
...
-		private static boolean hasMaxCount(Resource id, SailRepositoryConnection connection) {
-			return connection.hasStatement(id, SHACL.MAX_COUNT, null, true);
+		private static boolean hasMaxCount(Resource id, ShaclSailConnection connection) {
+			return connection.hasStatement(id, SHACL.MAX_COUNT, null, true, ShaclSail.SHACL_GRAPH);
 		}
 
-		private static boolean hasDatatype(Resource id, SailRepositoryConnection connection) {
-			return connection.hasStatement(id, SHACL.DATATYPE, null, true);
+		private static boolean hasDatatype(Resource id, ShaclSailConnection connection) {
+			return connection.hasStatement(id, SHACL.DATATYPE, null, true, ShaclSail.SHACL_GRAPH);
```
hmottestad: So we just have to remember to use the named graph, and that’s it. Seems decent. If you forget the named graph, will it return nothing?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -26,64 +37,57 @@
...
-
 /**
  * @author Heshan Jayasinghe
  */
 public class ShaclSailConnection extends NotifyingSailConnectionWrapper {
 
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 
-	private NotifyingSailConnection previousStateConnection;
+	//	private NotifyingSailConnection previousStateConnection;
```
hmottestad: What happened to the previousStateConnection?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -172,15 +175,21 @@ private boolean validate() {
 
 		boolean allValid = true;
 
-		for (Shape shape : sail.shapes) {
+		for (Shape shape : sail.getShapes(this)) {
```
hmottestad: So every commit reads in all the shapes again? Seems costly. Plans on using the notifying sail aspect to pick up changes to the shacl named graph and use some caching for the shapes?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -190,41 +199,69 @@ private boolean validate() {
...
-
+	
 	@Override
-	synchronized public void close() throws SailException {
+	public CloseableIteration<? extends Statement, SailException> getStatements(Resource subj, IRI pred,
+			Value obj, boolean includeInferred, Resource... contexts)
+		throws SailException
+	{
+		if (contexts.length == 0) {
+			List<Resource> graphs = new ArrayList<>();
```
hmottestad: I feel that this approach will add a lot of overhead to each query. Have you looked at the benchmarks?

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/ReduceNumberOfPlansTest.java
```diff
@@ -8,114 +8,126 @@
...
 
-	@Test
-	public void testAddingTypeStatement() {
-		SailRepository shaclSail = new SailRepository(new ShaclSail(new MemoryStore(), Utils.getSailRepository("reduceNumberOfPlansTest/shacl.ttl")));
-		shaclSail.initialize();
+	private SailRepository shaclRepo;
+
+	@Before
+	public void setup() {
+		shaclRepo = TestUtils.getShaclRepository("reduceNumberOfPlansTest/shacl.ttl");
```
hmottestad: This test is supposed to test with different shacl files to make sure that plans for rules that don’t need to be triggered are skipped. So this will need another rewrite to handle that anyway.

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/ShaclTest.java
```diff
@@ -35,13 +38,13 @@
 @RunWith(Parameterized.class)
 public class ShaclTest {
 
-	static final List<String> testCasePaths = Arrays.asList(
-		"test-cases/datatype/simple",
-		"test-cases/minCount/simple"
-	);
+	static final List<String> testCasePaths = Arrays.asList("test-cases/datatype/simple",
```
hmottestad: Can you multiline this again? Looks like eclipse is a bit aggressive here with the style guide. I expect this list to become quite long.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/DatatypePropertyShape.java
```diff
@@ -34,49 +35,52 @@
...
-
-		PlanNode top = new LoggingNode(new InnerJoin(bufferedSplitter.getPlanNode(), invalidValuesDirectOnPath, null, discardedRight));
-
+		PlanNode top = new LoggingNode(new InnerJoin(bufferedSplitter.getPlanNode(),
+				invalidValuesDirectOnPath, null, discardedRight));
 
 		if (shape instanceof TargetClass) {
-			PlanNode typeFilterPlan = new LoggingNode(((TargetClass) shape).getTypeFilterPlan(shaclSailConnection.getPreviousStateConnection(), discardedRight));
+			PlanNode typeFilterPlan = new LoggingNode(
+					((TargetClass)shape).getTypeFilterPlan(shaclSailConnection, discardedRight));
```
hmottestad: The use of previousStateConnection here is for performance reasons. In essence, at this point in the plan we only care about data type statements where the subject has a matching type, but we have eliminated those where the type statement was added in the current transaction. So we don’t have to query that data again.

hmottestad:
Hi Jeen.

Good work :) I like the use of a named graph, but a bit nervous about the potential performance issues. 

Håvard

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/DatatypePropertyShape.java
jeenbroekstra: The reason I changed this was that I am trying to get it to a state where it doesn't mix up the two APIs. `ShaclSailConnection` is a `SailConnection` (and therefore part of the SAIL API), while `SailRepositoryConnection` is a `RepositoryConnection` (and therefore part of the Repository API). I really dislike the fact that these are mixed together - they're APIs at different levels, with different purposes. Getting rid of the secondary repo for the rules means we can do away with this layer confusion altogether and handle most everything at the level of the SAIL.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
jeenbroekstra: That's the idea. Note that I am still working on that part: for example this will now work for `getStatements` calls, but not yet for SPARQL queries. For that, I will need to do some trickery with the injected `Dataset` object on a call to `SailConnection.evaluate`

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
jeenbroekstra: I took it out because I believe that with the switch to a single repo for rules and data, it's no longer necessary. To be fair I didn't clearly understand what it was trying to achieve, it may be that I overlooked something, but no other layering/reasoning SAIL needs this kind of construction, so I am operating under the assumption that we don't really need this. If I'm wrong, it's easy to put it back :)

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
jeenbroekstra: Actually it doesn't. The sail keeps a private `List<Shape>` that, if not null, is immediately returned. Only when that list is not yet populated does it retrieve the shapes from the DB. But yeah, a further improvement on that would be some listener mechanism that updates this list if the shapes graph gets changed.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
jeenbroekstra: I haven't yet, and you're right that is likely to be costly. But I want to get things functionally correct before I start optimizing too much. An easy optimization here would be to keep a private cached list of graph names (similar to how we cache the list of Shape objects).

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/ShaclTest.java
jeenbroekstra: I'll take a look. Eclipse formatting usually takes care of this as soon as it becomes too long - it only put it on a single line because right now, it fits. But I'll doublecheck.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/DatatypePropertyShape.java
jeenbroekstra: Given that we already keep track of all added and removed statements in the current connection (the `addedStatements` and `removedStatements` sets), couldn't we implement this same optimization using those lists instead of a separate connection?

hmottestad: It was meant to be used to reduce the amount of data that had to be retrieved. I can’t remember the exact case now, and I’m not sure if I have a benchmark for that case either. These cases are anyways only documentable by benchmarks.

hmottestad:
@jeenbroekstra ok, thanks for letting me know. I've wanted to implement sh:or and sh:and, to see if the overall code design is correct. However the biggest problem has been lack of time. July is often a quieter time at work, so maybe I'll find some time then.

jeenbroekstra:
I've left this way too long, it's gone pretty stale. I'll make an effort to update and finish this week.

hmottestad:
Any chance we could merge in the sh:or support first? 

https://github.com/eclipse/rdf4j-storage/pull/109

That one only has one thing missing to be spec compliant. Which is support for some sort of inheritance or inference to deduce the sh:path based on inheritance.

jeenbroekstra:
@hmottestad I'm fine with merging your PR in first when it's done. This refactor is not complete yet anyway, still need to resolve some major problems.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/DatatypePropertyShape.java.orig
```diff
@@ -0,0 +1,122 @@
...
+import java.util.stream.Stream;
+
+/**
+ * @author Håvard Ottestad
+ */
+public class DatatypePropertyShape extends PathPropertyShape {
+
+	private final Resource datatype;
+
+<<<<<<< HEAD
```
hmottestad: did you commit a file with a merge conflict?

jeenbroekstra: I did. Drat, good catch. Removed them.

barthanssens:
Perhaps org.apache.commons.**text**.StringEscapeUtils could be used of org.apache.commons.**lang**.StringEscapeUtils

(commons lang 2 is a legacy library, while recent commons lang 3 library deprecated lang.StringEscapeUtils in favor of text.StringEscapeUtils), see also #1173

jeenbroekstra:
I am closing this and am about to put up a new PR, with a more modest refactoring - aiming purely at removing references to the Repository API from the public API of this Sail

PR #72, opened 2018-05-23
Fix #71: Update Lucene, Solr and Elasticsearch dependencies

### LINKED ISSUE #71: Upgrade Lucene, Elasticsearch and Solr Dependency to Lucene 6.x
In the current RDF4J release Lucene 5.x is used as a basis for keyword search and search indexes. This version of lucene is outdated and declared Branch End-Of-Live.

The same holds for the elasticsearch sail (version 2.x) and solr.

For security reasons it is highly recommended to upgrade these 3rd party dependencies.

It is to be evaluated whether  the upgrade will be done in two steps or in a single step, i..e whether to skip a major lucene version (6.x) or not.

As a first step the codebase is to be updated to the following versions

Lucene (5.5.4 -> 6.6.4)
Elastic Search (2.4.6 -> 5.6.9)
Solr (5.5.4 -> 6.6.4)

Later (or in a next task) it may be decided to do the step to a lucene 7.x based dependency

aschwarte10:
This PR addresses GitHub issue: #71  .

This change updates outdated and no longer maintained versions of Lucene
and the modules using it.

Updates

Lucene: 5.5.4 -> 6.6.1
Solr: 5.5.4 -> 6.6.1
Elasticsearch: 2.4.6 -> 5.6.9

Requires updated of the following transitive dependencies

- spatial4j: 0.5 -> 0.6

### DIFF
--- a/elasticsearch/src/main/java/org/eclipse/rdf4j/sail/elasticsearch/ElasticsearchIndex.java
+++ b/elasticsearch/src/main/java/org/eclipse/rdf4j/sail/elasticsearch/ElasticsearchIndex.java
@@ -18,7 +18,6 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 
-import org.apache.lucene.spatial.util.GeoHashUtils;
 import org.eclipse.rdf4j.model.Resource;
 import org.eclipse.rdf4j.model.URI;
 import org.eclipse.rdf4j.model.vocabulary.GEOF;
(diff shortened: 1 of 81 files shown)


aschwarte10:
The change is work in progress, update of lucene and solr is looking good, elastic search is not yet done (problems in maven transitive dependencies cause the integration test to fail). Still looking into it

elasticsearch/src/main/java/org/eclipse/rdf4j/sail/elasticsearch/ElasticsearchIndex.java
```diff
@@ -253,7 +254,8 @@ else if ("yellow".equals(waitForStatus)) {
 		}
 		String waitForRelocatingShards = parameters.getProperty(WAIT_FOR_RELOCATING_SHARDS_KEY);
 		if (waitForRelocatingShards != null) {
-			healthReqBuilder.setWaitForRelocatingShards(Integer.parseInt(waitForRelocatingShards));
+			// TODO no longer available
+//			healthReqBuilder.setWaitForRelocatingShards(Integer.parseInt(waitForRelocatingShards));
```
aschwarte10: I could not find a way to migrate this to the new elasticsearch API. Do we need this?

jeenbroekstra: I found this: https://www.elastic.co/guide/en/elasticsearch/reference/5.3/breaking_50_java_api_changes.html#_clusterhealthrequestbuilder . Looks like they recommend replacing with `setWaitForNoRelocatingShards(boolean)`.

PR #80, opened 2018-06-22
Upgrade jackson to 2.9.6

aschwarte10:
This PR addresses GitHub issue: 75# .

Updates jackson to 2.9.6 to address a security vulnerability in jackson 2.9.5

See also https://github.com/eclipse/rdf4j/issues/1035


jeenbroekstra:
I just realized that I overlooked what you mentioned earlier, which is that we might be able to reuse dependency version settings from the main project. I'll have a quick look at the maven setup to see if we can do that. In the meantime, this looks fine to me though.

PR #83, opened 2018-06-23
Features/rdf4j#461project parent pom streamlining

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#461 .

* remove duplicate project config from `rdf4j-storage-parent` pom and instead inherit from `rdf4j-parent`
* add restlet repo to `solr-sail` pom since it's only needed there (for a transitive `solr-core` dependency)
* added `ZZ-` prefix to the repo id so it gets queried last of the known repos (ObDirtyHack)


jeenbroekstra:
Build fails because required change in rdf4j has not yet been merged, so it can't find the new parent pom.

PR #90, opened 2018-08-05
GH-89 include JTS and use as default SpatialSupport

jeenbroekstra:
This PR addresses GitHub issue: #89 .

* bumped spatial4j to 0.7
* included jts 1.15.1
* refactored SpatialSupport and related classes to use full JTS functionality by default

### DIFF
--- a/compliance/elasticsearch/pom.xml
+++ b/compliance/elasticsearch/pom.xml
@@ -34,7 +34,7 @@
 			</plugin>
 		</plugins>
 	</build>
-	
+
 	<dependencies>
 		<dependency>
 			<groupId>${project.groupId}</groupId>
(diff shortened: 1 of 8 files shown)


jeenbroekstra:
@pulquero I based some off this of your bitbucket project that demonstrates a SpatialAlgebra implementation using JTS, but updated to the latest spatial4j and jts APIs. Keen to hear what you think.

jeenbroekstra:
There's three GeoSPARQL compliance tests failing, it looks like it has something to do with inclusion of "empty" points, but I can't immediately see what's wrong.

pulquero:
Nice, I think the problem with the tests is to do with wkt serialization. An empty geometry is being written out as "GEOMETRYCOLLECTION ()"^^<http://www.opengis.net/ont/geosparql#wktLiteral> and not "GEOMETRYCOLLECTION EMPTY"^^<http://www.opengis.net/ont/geosparql#wktLiteral>. You probably need to modify toWKT() to something like if(!shape.isEmpty()) return wktWriter.toString(shape); else return "GEOMETRYCOLLECTION EMPTY";

jeenbroekstra:
I've logged two issues for possible Spatial4J bugs related to this:

https://github.com/locationtech/spatial4j/issues/162
https://github.com/locationtech/spatial4j/issues/163

For now I've managed to implement workarounds that at least cover our test cases.

Issue #95, opened 2018-08-13
Improve parallelism support in nativerdf


mirzov:
Discovered that parallel multithreaded execution of SPARQL queries on nativerdf result in blocking of threads and no increase in query throughput (in comparison with a single-threaded queued execution). Moreover, the execution time of each of the queries becomes the sum of execution times of each (in single-threaded mode). Additionally, discovered that parallel execution of queries using a ForkJoinPool thread pool does not work (exception is thrown). (Earlier, found that parallel multi-threaded ingestion of triples into different RDF graphs results in dead lock, but that may be unrelated).
Thread dump during parallel SPARQL execution showed that blocking occurs while waiting for synchronization locks on nodeCache in org.eclipse.rdf4j.sail.nativerdf.btree.BTree. The reason seems to be needlessly aggressive locking that is being done even during read operations.

PR #97, opened 2018-08-15
GH-95 Improving parallelism in nativerdf

mirzov:
This PR addresses GitHub issue: #95  .

The changes proposed in this PR:

* Getting rid of synchronization in `BTree` and `Node` (except on `allocatedNodesList`)
* Replacing the caches `LRUCache` and `nodeCache` and `mruCache` in `BTree` and `ValueStore` with alternative lock-free implementations based on `ConcurrentHashMap`. The old versions of the caches are removed in the last commit.
* Splitting the large BTree.java file into several, moving test code out.

Thread dumps during parallel execution of expensive SPARQL queries show that, as a result of this change, the threads do not any more spend most of their time blocked, waiting for locks. Test runs showed that the performance penalty for parallel execution of 5 expensive queries has been reduced from at least the factor of 5 to the factor of 2.

Additionally, performance of single queries appears to have also been improved. For example, one observed reduction in execution time was from 1.9 seconds to 1.4 seconds.


mirzov:
Dear maintainers @jeenbroekstra , @jamesrdf , @hmottestad ,

Please consider reviewing this pull request, despite the failing CI. The reason for failure is that it takes too much time to download the dependencies and the CI process is killed by Travis due to it taking too much time. All the tests pass locally (ran `mvn clean` followed by `mvn verify`). By the way, had to add the Sonatype snapshots repository to the main POM file for builds to work on the current master (not included in this pull request).

jeenbroekstra:
Thanks for this @mirzov . I'll try and take a closer look ASAP but it sounds promising!

mirzov:
Thanks for the approval @jeenbroekstra ! Spent more time thinking about the code and introduced a few more changes to make it more robust. Ran `mvn clean` and `mvn verify` on rdf4j-storage locally, all the tests still pass, and performance characteristics in my own experiments did not suffer, either. Please review the last commit when you have time. Do you think these changes can be merged for the coming 2.4.0 release?

jeenbroekstra:
I'll have a closer look tonight but I don't see why not. The one thing that worries me slightly is Travis timing out. I don't immediately see anything in your PR that would cause this, but other PRs don't have this issue.

nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/btree/BTree.java
```diff
@@ -97,7 +97,7 @@
 			node.read();
 		}
 		catch (IOException exc) {
-			logger.error("Error reading B-Tree node", exc);
+			throw new RuntimeException("Error reading B-tree node", exc);
```
jeenbroekstra: Can we change that to a `SailException`? It's an RDF4J-specific subclass of `RuntimeException`.

mirzov:
Sure thing! It is done. Also, Travis does not time out anymore; it must have been some temporary issue on their side.

jeenbroekstra:
Thanks again @mirzov , merged, and will be included in the 2.4 release.

Issue #99, opened 2018-09-09
Visualize query plan for SHACL execution


hmottestad:
Query plans are hard to grasp just from reading the code. Parts of the plan are printed in the logs when executed, but it's still hard to understand what's going on. Printing the raw nodes in the plan will help. Since there is a lot of interconnectivity it might be best to print them as GraphViz DOT.

PR #100, opened 2018-09-09
#99 visualize min count, max count and datatype

hmottestad:
This PR addresses GitHub issue: #99  .

* adds a method to sail for enabling debug plan printing
* adds methods to plan items to print GraphViz Dot format to stdout
* adds three tests that print output, but don't verify anything

### DIFF
--- a/shacl/pom.xml
+++ b/shacl/pom.xml
@@ -18,6 +18,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+        </dependency>
+        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>rdf4j-sail-api</artifactId>
             <version>${project.version}</version>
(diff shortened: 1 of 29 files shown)


hmottestad:
gsoc - shacl

hmottestad:
Above is an example output which I have annotated with some more explanations.

hmottestad:
Output from program can be pasted into a tool such as http://viz-js.com/ for visualization.

shacl/pom.xml
```diff
@@ -17,6 +17,10 @@
     <description></description>
 
     <dependencies>
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+        </dependency>
```
hmottestad: I saw that the SPIN module was already using this dep so I reckon it's fine if I use it too, IP wise.

hmottestad:
@jeenbroekstra I'm hoping this will help people in understanding what the java code does when evaluating the shacl rules, and it should help as we develop support for new rules and more optimisations.

hmottestad:
Since this is so small and for debug purposes, I’ll merge it on Monday if I don’t hear anything.

Issue #106, opened 2018-09-12
Support for sh:or


hmottestad:
Support for sh:or in both property shapes and node shapes.

jeenbroekstra:
This issue was moved to eclipse/rdf4j#1118

PR #109, opened 2018-09-25
Feature/106 sh or

hmottestad:
This PR addresses GitHub issue: #106  .

* support for sh:or
* 
*


hmottestad:
Currently missing some sort of inference for inheriting sh:path.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
```diff
@@ -59,28 +60,32 @@ public boolean requiresEvaluation(Repository addedStatements, Repository removed
 		return false;
 	}
 
+
+	public void printPlan(PlanNode planNode, ShaclSailConnection shaclSailConnection) {
+
+		System.out.println("digraph  {");
```
jeenbroekstra: Shouldn't this be logged somewhere instead of printed to STDOUT?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/DatatypePropertyShape.java
```diff
@@ -48,7 +47,7 @@
 
 
 	@Override
-	public PlanNode getPlan(ShaclSailConnection shaclSailConnection, NodeShape nodeShape) {
+	public PlanNode getPlan(ShaclSailConnection shaclSailConnection, NodeShape nodeShape, boolean printPlans, boolean assumeBaseSailValid) {
```
jeenbroekstra: I haven't completely grokked what this parameter `assumeBaseSailValid` does, but given that you have a reference to the ShaclSailConnection (and therefore to the Sail), perhaps we could make this a boolean method on the Sail object instead of having to inject it as a param everywhere?

hmottestad: Could look at placing it somewhere else. I just recently had to introduce it. 

The reason for introducing it is that up until now, the implementation has always assumed that the data currently in the base sail is valid. But when you have sh:or, for instance sh:or (cellphoneNumber min 1 OR landlineNumber min 1), then either one of those sub rules can be invalid in the base sail. So we need to tell the minCount implementation to de-optimize it’s validation plan. 

At the moment, you kindof inherit assumeBaseSailValid from your “parent” rule or shape. With some rules overriding the value, eg sh:or doesn’t care if it’s wrapped in another sh:or or sh:and.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
hmottestad: Should probably be collected by a string builder. 

Point is to be able to copy paste the output into a graphiz visualizer online. So if each line were logged you would have date time, thread name, etc. per line to clean up.

hmottestad:
I can clean up the merge conflicts tonight.

hmottestad:
Fixed the conflicts, minor stuff.

Hoping the build completes.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
jeenbroekstra: Ah I hadn't  cottoned on that this was input for GraphViz.

I'm still not comfortable with just hardcoded printing to STDOUT though. Like you say, a stringbuilder that can get redirected and/or a customized plan-log file (where you have full control over what it looks like) seems a better approach.  

By the way: the "plan" that is being printed here, what exactly is it? Are we talking an execution plan for a specific SHACL rule or something?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/DatatypePropertyShape.java
jeenbroekstra: We can leave this for later if you prefer, I can look at it as part of the refactoring once this PR is complete and merged

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
hmottestad: I like to call it the validation plan. It’s actually a query plan, that returns invalid data.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/DatatypePropertyShape.java
hmottestad: Probably best to leave it as is until the other logical operators are implemented. Sh:and, sh:not and sh:xone.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
hmottestad: Btw. This is just extracting a method. Was also println in my original code (on develop). I’ll make a new issue for this.

hmottestad: https://github.com/eclipse/rdf4j/issues/1155

PR #110, opened 2018-10-01
eclipse/rdf4j#1018 fix handling of null binding in iteration

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#1018 .

* QueryBindingSet iterator filters out any bindings with a `null` special value (see eclipse/rdf4j#79)
* added regression test

### DIFF
--- a/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
+++ b/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
@@ -124,7 +124,9 @@ public class QueryBindingSet extends AbstractBindingSet {
 	}
 
 	public Iterator<Binding> iterator() {
-		Iterator<Map.Entry<String, Value>> entries = bindings.entrySet().iterator();
+
+		Iterator<Map.Entry<String, Value>> entries = bindings.entrySet().stream().filter(
+				entry -> entry.getValue() != null).iterator();
 
 		return new ConvertingIterator<Map.Entry<String, Value>, Binding>(entries) {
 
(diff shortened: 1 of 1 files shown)


evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
```diff
@@ -124,7 +124,9 @@ public boolean hasBinding(String bindingName) {
 	}
 
 	public Iterator<Binding> iterator() {
-		Iterator<Map.Entry<String, Value>> entries = bindings.entrySet().iterator();
+
+		Iterator<Map.Entry<String, Value>> entries = bindings.entrySet().stream().filter(
+				entry -> entry.getValue() != null).iterator();
```
hmottestad: https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#nonNull-java.lang.Object-

Objects::nonNull

hmottestad:
Looks fine. Recommended some fancy slick java 8 stuff ;) btw. where do the nulls come from?

jeenbroekstra:
@hmottestad it's used as a marker for unbound vars that should not be re-evaluated. For example when a type error occurs inside a BIND clause. See #79. TBH I'm not keen on using null this way, but it works for now.

jeenbroekstra:
Er that's eclipse/rdf4j#79

evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/QueryBindingSet.java
jeenbroekstra: @hmottestad hadn't thought of that, but I'm not immediately sure how I'd use that effectively here. Note that I'm not checking if the object in the stream itself is null, but whether one of its methods returns null.

PR #111, opened 2018-10-04
 Merge 2.4.0 release master into develop

barthanssens:
Sync develop branch with small changes in release 2.4.0 master (without touching the versions in pom.xml, which are already set to 2.5-SNAPSHOT)


jeenbroekstra:
Thanks for this Bart. The pom version issue is not something that should occur if I do my job right, but with all the milestones and the long release process things got a bit messy.

PR #78, opened 2018-06-07
eclipse/rdf4j#1114 Upgrade to Lucene 7.x based indices

aschwarte10:
This PR addresses GitHub issue: eclipse/rdf4j#1114  .

Upgrade components

Lucene: 6.6.1 -> 7.2.1
Elasticsearch: 5.6.9 -> 6.2.4
Solr: 6.6.1 -> 7.2.1

Note: we cannot upgrade to lucene 7.3.1 as elasticsearch has a
strict version dependency on 7.2.1. Also spatial4j cannot seamlessly be
upgraded

PR #124, opened 2018-10-29
eclipse/rdf4j#78 fix join optimizer messing with bind/subselect combo

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#78 .

* ensure QueryJoinOptimizer does not mess up order when a BIND clause is combined with a subselect
* added regression test

### DIFF
--- a/evaluation/pom.xml
+++ b/evaluation/pom.xml
@@ -59,6 +59,7 @@
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
+			<scope>test</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
@barthanssens if possible I'd like this merged in before the 2.4.1 release.

jeenbroekstra:
Retracting for now, this needs a closer look I think, as  I may have overlooked some edge cases

jeenbroekstra:
I had a closer look and the edge case I had in mind is not a problem, as that case (different order of subselect and bind) does not actually involve a Join.

PR #131, opened 2018-11-23
Feature/1118 more sh or support

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1118 .

* more debug printing of execution plan
* support for short-hand syntax for sh:or by expanding with SPARQL


shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/planNodes/EqualsJoin.java
```diff
@@ -57,7 +59,11 @@ void calculateNext() {
 					if (nextRight != null) {
 
 						if (nextLeft.line == nextRight.line || nextLeft.line.equals(nextRight.line)) {
-							next = TupleHelper.join(nextLeft, nextRight);
+							if(useAsFilter){
+								next = nextLeft;
+							}else{
+								next = TupleHelper.join(nextLeft, nextRight);
+							}
```
hmottestad: EqualsJoin joins equal tuples, eg "A, B, C" can be joined with "A, B, C" but not with "A, B, C, D" or "A, B, E".

I use EqualsJoin in sh:or to filter tuples that are equal, so I don't want the joined result.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/planNodes/FilterPlanNode.java
```diff
@@ -74,11 +75,19 @@ private void calculateNext() {
 
 					if (checkTuple(temp)) {
 						if (trueNode != null) {
+							if(LoggingNode.loggingEnabled){
+								System.out.println(leadingSpace() + that.getClass().getSimpleName() + ";trueNode: " + " " + temp.toString());
```
hmottestad: I know this should use the logger, but I am currently using stdout everywhere else for this debug logging...so I will instead make a new PR to clean that up all in one go later.

PR #132, opened 2018-12-09
Replace some REGEX calls that can be expressed as string functions

JervenBolleman:
Addresses issue: eclipse/rdf4j#1199

This is a trivial query optimizer which can help public endpoints where cheaper string functions (that are easier to further optimize) instead of relatively expensive regex calls. (even if the regex is otherwise trivial in runtime costs).

The regex pattern is tested to make sure it does not contain any of the control characters and is case sensitive. This could be improved to avoid false negatives by inspecting the Pattern object itself. However,
that needs reflection and I think that is not welcome.

This pull request just adds the optimizer, it does not enable it any repository.

There is an copyright assignment from me (and my employer) to Aduna on file. Arranging a new one for RDF4J will take a lot of time, so I hope this acceptable as is.

### DIFF
--- a/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizer.java
+++ b/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizer.java
@@ -1,0 +1,116 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Eclipse RDF4J contributors.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *******************************************************************************/
+package org.eclipse.rdf4j.query.algebra.evaluation.impl;
+
+import org.eclipse.rdf4j.model.ValueFactory;
+import org.eclipse.rdf4j.model.vocabulary.FN;
+import org.eclipse.rdf4j.query.BindingSet;
+import org.eclipse.rdf4j.query.Dataset;
+import org.eclipse.rdf4j.query.algebra.Compare;
+import org.eclipse.rdf4j.query.algebra.FunctionCall;
+import org.eclipse.rdf4j.query.algebra.Regex;
+import org.eclipse.rdf4j.query.algebra.TupleExpr;
+import org.eclipse.rdf4j.query.algebra.ValueConstant;
+import org.eclipse.rdf4j.query.algebra.ValueExpr;
+import org.eclipse.rdf4j.query.algebra.Var;
+import org.eclipse.rdf4j.query.algebra.evaluation.QueryOptimizer;
+import org.eclipse.rdf4j.query.algebra.helpers.AbstractQueryModelVisitor;
+
+/**
+ * A query optimizer that replaces REGEX with {@link FunctionCall}s that are
+ * equivalent operators
+ *
+ * @author Jerven Bolleman
+ */
+public class RegexAsStringFunctionOptimizer implements QueryOptimizer {
+
+    private final ValueFactory vf;
+
+    public RegexAsStringFunctionOptimizer(ValueFactory vf) {
+        this.vf = vf;
+    }
+
+    /**
+     * Applies generally applicable optimizations to the supplied query:
+     * variable assignments are inlined.
+     */
+    public void optimize(TupleExpr tupleExpr, Dataset dataset, BindingSet bindings) {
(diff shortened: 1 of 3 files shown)


jeenbroekstra:
Could you please digitally sign the Eclipse Contributor Agreement (ECA)? See the [Contributor Guidelines](https://github.com/eclipse/rdf4j-doc/blob/master/.github/CONTRIBUTING.md) for more details on how to sign. Unfortunately, we cannot accept your contribution until you have signed this. It shouldn't take more than 5 minutes.

evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizer.java
```diff
@@ -0,0 +1,116 @@
+/** *****************************************************************************
+ * Copyright (c) 2018 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: This is the wrong license header. Please use 

```
/*******************************************************************************
 * Copyright (c) 2018 Eclipse RDF4J contributors.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Distribution License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/org/documents/edl-v10.php.
 *******************************************************************************/
```

evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizierTest.java
```diff
@@ -0,0 +1,132 @@
+/** *****************************************************************************
+ * Copyright (c) 2018 Eclipse RDF4J contributors, Aduna, and others.
```
jeenbroekstra: Ditto here.

jeenbroekstra:
We'll need to rebase this PR to the develop branch before merging (I tried just now but it messes up the diff a bit, so decided to undo until we're good to go with the merge)

JervenBolleman:
I am struggling with the eclipse account signup :( but will do so.

jeenbroekstra:
@JervenBolleman I've noticed the Eclipse site was a bit slow and flaky today. Let me know if you hit a wall and I'll try and get a maintainer involved.

JervenBolleman:
ECA is signed, contribution SOP inside my workplace followed and not objected to so allowed to make the contribution.

Linking eclipse account to github id does not seem to work nicely so you might need to do some manual work to link it.

JervenBolleman:
The e-mail is jerven.bolleman@sib.swiss

jeenbroekstra:
Thanks, confirmed!

evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizierTest.java
jeenbroekstra:
```suggestion
 * Copyright (c) 2018 Eclipse RDF4J contributors.
```

JervenBolleman:
Sorry for missing the second license header.
Is the content of the patch itself ok? 
No regexes that could be converted when they shouldn't be?

jeenbroekstra:
Thanks for fixing @JervenBolleman . I would normally just correct such small things myself but for some reason your PR is not editable for me.

 Anyway: content looks ok to me, you've got a fair spread of test cases. A more comprehensive test would be to run a full SPARQL WG testsuite with this optimizer enabled, and see what happens. In fact it might be a good idea to do that as part of this PR, if you're up for it. 

We should also have a think about whether we want this enabled in the core stores (memory and native).

jeenbroekstra:
@rdf4j-bot run tests

JervenBolleman:
I am ok to run the full test suite. I think it makes sense to enable it for the native and memory stores. The evaluation time to determine if this optimizer should run is minimal and not enabled code tends to bitrot.
While gains are minimal so are costs.

In any case I started with this minimal optimizer before starting on more complicated ones to get familiar with the process.

jeenbroekstra:
@JervenBolleman are you able to update this PR to enable this optimizer in the Memory- and NativeStore by default? That should also automatically tell us if we've overlooked any corner cases, as it will then be used in all our compliance tests.

JervenBolleman:
@rdf4j-bot run tests

@jeenbroekstra sorry between travel and christmas I misunderstood what you liked. I think the last commit should turn everything on.

JervenBolleman:
Sorry for force pushing but that fixes the ECA check...

jeenbroekstra:
@rdf4j-bot run tests

PR #133, opened 2018-12-09
eclipse/rdf4j#1197 add statements with correct graph

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1197 .

* inferred statements get graph based on added statement
* 
*

### DIFF
--- a/inferencer/src/main/java/org/eclipse/rdf4j/sail/inferencer/fc/SchemaCachingRDFSInferencer.java
+++ b/inferencer/src/main/java/org/eclipse/rdf4j/sail/inferencer/fc/SchemaCachingRDFSInferencer.java
@@ -93,6 +93,13 @@ public class SchemaCachingRDFSInferencer extends NotifyingSailWrapper {
 	// The previous transaction rolled back
 	boolean rolledBackAfterModifyingSchemaCache;
 
+	// Inferred statements can either be added to the default context
+	// or to the context that the original inserted statement has
+	// for the time being, the default behaviour will be to adde the
+	// statements to the default context.
+	// THIS BEHAVIOUR WILL BE SWITCHED ON THE NEXT MAJOR RELEASE
+	private boolean addInferredStatementsToDefaultContext = true;
+
 	/**
 	 * Instantiate a SchemaCachingRDFSInferencer.
 	 *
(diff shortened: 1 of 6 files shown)


hmottestad:
Need to merge the testfixes first: https://github.com/eclipse/rdf4j-testsuite/pull/40

hmottestad:
@barthanssens @jeenbroekstra 

Had to combine multiple RDF files into one by having each file in a graph. Then used the ContextAwareRepository to read/write a single graph. Essentially getting a virtual repository that looked like everything was in the default graph, but instead it was just projecting a graph from the underlying sail which contained one graph per file. 

Found out that when doing this reasoning wasn't working as expected. So ended up having the RDFS reasoner adding the inferred statements to the context (graph) based on the context of the added statement. 

Do you guys think this makes sense to do?

hmottestad:
@jeenbroekstra I've added a ```sail.setAddInferredStatementsToDefaultContext(boolean)``` method to the sail and set it to true as default.

If I remember correctly merging this to master makes it a minor release and then I can make a new PR with default as false and merge that into develop which will result in a major release? Or was it the other way around?

jeenbroekstra:
@hmottestad merging to master makes it part of the next _patch_ release, merging to develop is for the next _minor_ release. We currently have no branch specifically for the next major release. 

So I think this needs to be merged to develop. As for a followup PR to make a change that needs a major release: it may be time for us to start a branch for development of a 3.0 release. Not to mention start planning around when we want to release this, and what should be part of it.

jeenbroekstra:
No - the definition of a minor release is that it contains only backward compatible/non-breaking changes :)

inferencer/src/main/java/org/eclipse/rdf4j/sail/inferencer/fc/SchemaCachingRDFSInferencer.java
```diff
@@ -93,6 +93,13 @@
 	// The previous transaction rolled back
 	boolean rolledBackAfterModifyingSchemaCache;
 
+	// Inferred statements can either be added to the default context
+	// or to the context that the original inserted statement has
+	// for the time being, the default behaviour will be to adde the
+	// statements to the default context.
+	// THIS BEHAVIOUR WILL BE SWITCHED ON THE NEXT MAJOR RELEASE
```
jeenbroekstra: Can you log a new issue for this so we don't forget to pick this up for the 3.0 milestone?

PR #140, opened 2018-12-24
fix reactor order and include shacl in shaded jar

jeenbroekstra:

* fixes shacl javadoc missing from SDK distro javadoc


hmottestad:
And thanks for figuring it out :)

PR #143, opened 2018-12-30
GH-1225 marked inferencer deprecated

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#1225 .

* deprecated `ForwardChainingRDFSInferencer` and related code.

Note that the SchemaCachingInferencer currently has no corresponding Config and Factory classes yet. They're easy to add though,


hmottestad:
You mean config for which rules to apply? The schema caching one had a Boolean in its constructor for all/most rules. Nothing more detailed than that. 

I think it has an embedded factory, for the fast instatiation part.

jeenbroekstra:
@hmottestad no I mean subclasses of `SailFactory` and of `SailImplConfig`. These are necessary to be able to instantiate the inferencer via the `RepositoryManager` (which is what the RDF4J Server and Console use to spin up Sail stacks from a serialized config file).

PR #146, opened 2019-01-13
eclipse/rdf4j#1113 shacl sail api refactoring

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#1113 .

* SHACL shape data internally stored in in-memory repository (persisted)
* public constructor no longer exposes reference to Repository API
* shape data is added by using `ShaclSail.SHAPE_GRAPH` as the context identifier

This is WIP - several tests currently do not succeed. @hmottestad please review. This needs to be stabilized ASAP for inclusion in 2.5 so any tips you have on things I've overlooked that make the current tests fail would be useful. 

This is a far more modest refactoring than the older one, by the way: for now, we can keep using SailRepository _internally_ in this Sail implementation. The main goal is to make sure that the public api of this Sail is consistent with how other Sails work.


hmottestad:
I’ll have a look tomorrow, maybe the day after.

jeenbroekstra:
I've pushed updating the NodeShapes into the ShaclSailConnection, and this makes a lot of the tests now succeed. Still getting 22 test failures though, all in the `ShaclTest` (which I think executes the W3C testsuite?)

jeenbroekstra:
Also: we talked about this before but I must admit I still don't understand what the `previousStateConnection` is for?

jeenbroekstra:
Also-also: there is this method `runInferencingSparqlQueries` which executes two SPARQL updates. In the original setup this was executed once upon initialization. Can you explain what this is supposed to do, and if it is something that needs to happen only once or every time the NodeShape data gets modified?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -8,65 +8,119 @@
...
+							"shacl-sparql-inference/sh_or_node_shape.rq"),
+					"UTF-8");
+		}
+		catch (IOException e) {
 			throw new IllegalStateException(e);
 		}
 
 	}
 
-	public ShaclSail(NotifyingSail baseSail, SailRepository shaclSail) {
```
hmottestad: Would be very nice with a bunch of constructors that take SHACL rules in various formats. To simplify usage. For instance one that takes a file and format, and another that takes a repo, and maybe one that takes a model.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -98,17 +112,17 @@ public void begin() throws SailException {
...
 		assert addedStatements == null;
 		assert removedStatements == null;
 
 		stats = new Stats();
 
 		// start two transactions, synchronize on underlying sail so that we get two transactions immediatly successivley
-		synchronized (sail){
+		synchronized (sail) {
 			super.begin(level);
+			shapesConnection.begin(IsolationLevels.SERIALIZABLE);
```
hmottestad: How does this affect performance if you run up a few threads?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -133,20 +146,49 @@ public void commit()
...
+			}
+			finally {
 				cleanup();
 			}
 		}
 	}
 
+	@Override
+	public void addStatement(Resource subj, IRI pred, Value obj, Resource... contexts) throws SailException {
+		if (contexts.length == 1 && contexts[0].equals(ShaclSail.SHAPE_GRAPH)) {
```
hmottestad: Looks like a simple and fast way of handling updates. From the bug in the rdfs reasoner. Do we need to implement the add and remove methods that take update contexts too?

hmottestad: Maybe, for performance, we could have a “make sure there is a transaction” method here. So that the transaction against the SHACL rules repo connection only starts if we try to update it?

hmottestad:
> I've pushed updating the NodeShapes into the ShaclSailConnection, and this makes a lot of the tests now succeed. Still getting 22 test failures though, all in the `ShaclTest` (which I think executes the W3C testsuite?)

I don't think the W3C test suite is merged in yet. We are missing a lot of SHACL functionality.

hmottestad:
> Also: we talked about this before but I must admit I still don't understand what the `previousStateConnection` is for?

We have access to four sources of data during validation: 
 - added statements
 - removed statements
 - current sail state
 - previous sail state

Previous sail state can be used when we want to enrich our data in a validation query without duplicating up our data. 

An example is:

 - Person age max 1

We add:
 - Peter age 1.
 - Peter a Person.

Then we can use the following steps in our validation:

 - get join of "age" and " a Person"
 - count
 - if count > 1 => fail
 - else, for each subject get age from previousConnection
 - count again

The same could be accomplished with just the current sail state:

  - get join of "age" and " a Person"
 - count
 - if count > 1 => fail
 - else, 
 -- trim to subject
 -- for each subject get age from currentConnection (or whatever that variable is called)
 -- count

The latter is the actual plan that is implemented. The previousConnection is actually only used in one edge case at the moment. Though I have started to see some more use for it in my attempt to implement support for sh:class (eg. all foaf:knows must point to something that is a foaf:Person).

hmottestad:
> Also-also: there is this method `runInferencingSparqlQueries` which executes two SPARQL updates. In the original setup this was executed once upon initialization. Can you explain what this is supposed to do, and if it is something that needs to happen only once or every time the NodeShape data gets modified?

It's literally forward chaining to generate new statements from the shapes. Specifically implemented to support that SHACL has multiple ways of representing sh:path when using sh:or.

The two SPARQL queries that are run in the inferencer basically "standardise" how sh:path is used, so that we don't have to change the java code that parses it. 

A big problem with the inferencer is that it adds the inferred statements directly. So if you modify the initial data, the inferred data doesn't get removed. 

A quick and dirty solution for this would be to add the inferred statements to a different named graph. And clear that and rerun the inferencer before refreshing the shapes.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra: I'd prefer utility classes at the level of the Repository API for this kind of stuff, not constructors. The SAIL is not intended to be directly accessed by end-users and I'm wary of adding this kind of "convenience" functionality at this level.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
jeenbroekstra: I haven't measured, but it shouldn't have much influence on performance even on multiple threads. It's a bog-standard memory store so it does optimistic locking, and any way the txn only gets used if you modify the shape data itself. FWIW I put this level in on a whim, we could probably get away with going down to READ_COMMITTED.

jeenbroekstra:
> Previous sail state can be used when we want to enrich our data in a validation query without duplicating up our data.
> 
> An example is:
> 
>     Person age max 1
> 
> We add:
> 
>     Peter age 1.
>     Peter a Person.
> 
> Then we can use the following steps in our validation:
> 
>     get join of "age" and " a Person" from addedStatements
>     count
>     if count > 1 => fail
>     else, for each subject get age from previousConnection
>     count again

I feel like I'm missing something. I understand that you're first evaluating against the new statements only as a performance boost (fail fast). But what advantage do you get from using `previousConnection` for the second part, rather than just the current underlying connection?

hmottestad:
> I feel like I'm missing something. I understand that you're first evaluating against the new statements only as a performance boost (fail fast). But what advantage do you get from using `previousConnection` for the second part, rather than just the current underlying connection?

In the case where a new person is added it doesn't pull any data from the underlying store.

Eg.
Start with empty store.
Add :Peter a :Person.
Add :Peter :age 1.

The previousConnection would return <empty> for :Peter :age ?age.

While the currentConnection would return <1> for :Peter :age ?age.

So using the previousConnection is more efficient. And even more so when adding a lot of data.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
hmottestad: Discoverability is important for developers. Plenty of developers who will be instantiating this directly. Requiring the shapes to be added after instantiation to a special named graph means that developers have to read the documentation to use the shacl sail. I reckon there will be plenty of people who will want to just use this as a quick "validate my data" rather than as a long lived sail.

Could we at least have a couple methods like:
 .setShapes(Repository/Model/File/InputStream shapes)

OR

ShaclSail.Factory.fromShapes(Sail baseSail, Repository/Model/File/InputStream shapes)

jeenbroekstra:
I'm sure you have figures to back this up but I must say it is surprising to me that this would make a significant difference. Especially since you can iterate lazily and just cut off when you've reached the limit needed  to see if the constraint is still valid or not. 

Anyway, we're getting a bit far afield here. Thanks for taking the time to talk me through it. We can revisit the pros and cons of various optimisations later - let's first get the public API in good shape.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
hmottestad: ok, a ShaclValidator might not be too bad.

hmottestad:
Might not be much of a difference. Haven't benchmarked it extensively. 

The abstractions in the code can't handle quite that level of laziness. It'll do a "group by count" and check the count. And it'll lazily get the data per count, which works because all the streams are sorted. However you can't tell the group by count to stop at a certain count.

hmottestad:
@jeenbroekstra I have updated your code to fix the inference aspect. Essentially the issue was that the refreshShapes method wasn't called prior to validation. I've also added some limitations.

1.  you can't modify shapes once they are loaded, throws an exception
1.  you can't validate data without first loading shapes, throws an exception

To support modifying shapes we need to do a bit more work. The inferencing queries run on the loaded shapes both add and remove triples. This is because SHACL has a short hand notation for sh:or which we don't want to support in the java code. Expanding this means generating new triples and removing the triples for the short hand notation. I also think we should have some tests to decide how isolation and serialization should work when updating shapes. At the moment I don't think there is any isolation! 

Point 2 is there to "teach" people how to use the SHACL engine now that it is no longer obvious how to load shapes.

hmottestad:
@jeenbroekstra 
Here is the current build failure:

```
[INFO] RDF4J SHACL compliance tests 2.5-SNAPSHOT .......... FAILURE [  3.215 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 33:22 min
[INFO] Finished at: 2019-01-16T05:46:33-05:00
[INFO] ------------------------------------------------------------------------
Waiting for Jenkins to finish collecting data
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:testCompile (default-testCompile) on project rdf4j-shacl-compliance: Compilation failure
[ERROR] Picked up JAVA_TOOL_OPTIONS: -Xmx2G -Djava.io.tmpdir=/tmp/genie.rdf4j
[ERROR] /home/hudson/genie.rdf4j/.jenkins/jobs/rdf4j-storage-verify-pullrequest/workspace/compliance/shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/SHACLComplianceTest.java:[97,9] error: constructor ShaclSail in class ShaclSail cannot be applied to given types;
[ERROR] required: NotifyingSail
[ERROR] found: NotifyingSail,SailRepository
[ERROR] reason: actual and formal argument lists differ in length
```

Essentially there is a compliance test in the other repo that you need to update to match the new way of loading shacl shapes.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
jeenbroekstra: - [ ] Yes on adding a variant for the UpdateContext, good point.
- [ ] lazy txn start sounds like a simple enough improvement.

jeenbroekstra:
@hmottestad great work, thanks! Yeah, I was aware of the compliance test but wanted to get the basics working. I'll pick that up ASAP. A further point is probably (like with the SchemaCachingInferencer) adding an SPI factory + config so we can use this sail as part of a stack. And of course you're right that we'll need to document.

I also want to move the graph identifier to a more prominent location, probably somewere in the model.vocabulary package. I think the time has come to mint a RDF4J vocabulary, as the successor to the SESAME namespace :)

PR #152, opened 2019-01-19
Issues/1249 support prepare in shaclsail

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1249 .

* use prepare for validation
* block changes after prepare has been run

### DIFF
--- a/compliance/store/src/test/java/org/eclipse/rdf4j/sail/memory/ShaclConcurrencyTest.java
+++ b/compliance/store/src/test/java/org/eclipse/rdf4j/sail/memory/ShaclConcurrencyTest.java
@@ -5,14 +5,12 @@
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/org/documents/edl-v10.php.
  *******************************************************************************/
- 
+
 package org.eclipse.rdf4j.sail.memory;
 
-import org.eclipse.rdf4j.repository.sail.SailRepository;
 import org.eclipse.rdf4j.sail.Sail;
 import org.eclipse.rdf4j.sail.SailConcurrencyTest;
 import org.eclipse.rdf4j.sail.SailException;
-import org.eclipse.rdf4j.sail.inferencer.fc.SchemaCachingRDFSInferencer;
 import org.eclipse.rdf4j.sail.shacl.ShaclSail;
 import org.junit.Ignore;
 import org.junit.Test;
@@ -24,20 +22,14 @@ public class ShaclConcurrencyTest extends SailConcurrencyTest {
 	 *---------*/
 
 	@Override
-	protected Sail createSail()
-		throws SailException
-	{
-		SailRepository shaclShapes = new SailRepository(new MemoryStore());
-		shaclShapes.initialize();
-		return new ShaclSail(new MemoryStore(), shaclShapes);
+	protected Sail createSail() throws SailException {
+		return new ShaclSail(new MemoryStore());
 	}
 
 	@Ignore
 	@Test
 	@Override
-	public void testConcurrentAddLargeTxnRollback()
-		throws Exception
-	{
+	public void testConcurrentAddLargeTxnRollback() throws Exception {
 		// empty since this test is ignored
 	}
 }
--- a/compliance/store/src/test/java/org/eclipse/rdf4j/sail/memory/ShaclConcurrencyTest.java
+++ b/compliance/store/src/test/java/org/eclipse/rdf4j/sail/memory/ShaclConcurrencyTest.java
@@ -1,0 +1,35 @@
+package org.eclipse.rdf4j.sail.memory;
+
+import org.eclipse.rdf4j.repository.sail.SailRepository;
+import org.eclipse.rdf4j.sail.Sail;
+import org.eclipse.rdf4j.sail.SailConcurrencyTest;
+import org.eclipse.rdf4j.sail.SailException;
+import org.eclipse.rdf4j.sail.inferencer.fc.SchemaCachingRDFSInferencer;
+import org.eclipse.rdf4j.sail.shacl.ShaclSail;
+import org.junit.Ignore;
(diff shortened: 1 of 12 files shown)


shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -227,6 +233,28 @@ synchronized public void close() throws SailException {
...
+		preparedHasRun = true;
+		super.prepare();
+
+		synchronized (sail) {
+			boolean valid = true;
+			try {
+				valid = validate();
+
+				if (!valid) {
+					rollback();
```
jeenbroekstra: I don't think you should call `rollback()` here. That's up to the user.

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/PrepareCommitTest.java
```diff
@@ -0,0 +1,88 @@
+package org.eclipse.rdf4j.sail.shacl;
+
+import org.eclipse.rdf4j.model.vocabulary.RDFS;
+import org.eclipse.rdf4j.sail.NotifyingSailConnection;
+import org.eclipse.rdf4j.sail.memory.MemoryStore;
+import org.junit.Test;
+
+public class PrepareCommitTest {
+
+	@Test(expected = Throwable.class)
```
jeenbroekstra: Would be good to be more specific about the exact expected exception.

jeenbroekstra: And, for what it's worth: most base sail implementations allow further changes after a call to `prepare` - though I have yet to figure out if this is by design or simply a bug.

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/PrepareCommitTest.java
```diff
@@ -0,0 +1,88 @@
+package org.eclipse.rdf4j.sail.shacl;
```
jeenbroekstra:
```suggestion
/*******************************************************************************
 * Copyright (c) 2019 Eclipse RDF4J contributors.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Distribution License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/org/documents/edl-v10.php.
 *******************************************************************************/
 
package org.eclipse.rdf4j.sail.shacl;
```

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
hmottestad: What option does the user have? Does rollback do more than cancel the transaction and cleanup the store? If a user doesn’t rollback and just closes the store (and we don’t call rollback in close), is that bad in any way?

compliance/store/src/test/java/org/eclipse/rdf4j/sail/memory/ShaclConcurrencyTest.java
```diff
@@ -0,0 +1,35 @@
...
+	{
+		SailRepository shaclShapes = new SailRepository(new MemoryStore());
+		shaclShapes.initialize();
+		return new ShaclSail(new MemoryStore(), shaclShapes);
+	}
+
+	@Ignore
+	@Test
+	@Override
+	public void testConcurrentAddLargeTxnRollback()
```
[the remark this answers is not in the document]
hmottestad: It was actually ignored in the file I copied from :P

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
jeenbroekstra: `close()` takes care of doing a `rollback()` if necessary, so that's taken care of. But the idea is that `prepare()` should not make things irreversible. For example, if `prepare()` fails not because of a SHACL validation issue but (for example) lack of disk space, a user should be able to do the following:

1. prepare -> exception (disk full) 
2. run a cleanup process (e.g. remove temp files)
3. prepare again -> success
4. commit

You can't do that if you choose to auto-invoke `rollback()`.

hmottestad: That's a really good example @jeenbroekstra 

At the moment I ```rollback()``` inside ```commit()``` if the transaction fails. I guess that's wrong too, should probably defer that to ```close()``` as well.

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/PrepareCommitTest.java
hmottestad: I'm going to stick with throwing an exception if changes are made after a prepare(). It does say in the documentation that your only allowed to call prepare(), rollback(), commit() or potentially close() after prepare().

hmottestad:
@jeenbroekstra I think this is all done now and ready to merge.

I've removed automatic rollback from everything except close(). And added some tests to check that that works as intended.

I'm sticking with throwing an exception when you attempt to change the sail after calling prepare(), it makes sense according to the documentation and also makes validation a little bit simpler.

jeenbroekstra:
There appears to be a compilation failure in one of the shacl compliance tests on this PR, looks like it hasn't caught on that the shaclsail constructor changed.

hmottestad:
Looks like Jenkins might be stuck on Running org.eclipse.rdf4j.sail.shacl.ShaclIsolationLevelTest

That’s not very promising.

hmottestad:
Restarted Jenkins build. Same issue.

PR #157, opened 2019-01-22
Issues/1253 shacl range

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1253 .

* Support for sh:minInclusive (<=) and the three other variants


shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/planNodes/LiteralComparatorFilter.java
```diff
@@ -0,0 +1,85 @@
...
+
+	@Override
+	boolean checkTuple(Tuple t) {
+		Value literal = t.line.get(1);
+
+		if(literal instanceof Literal){
+
+			IRI datatype = ((Literal) literal).getDatatype();
+
+			if(datatypesMatch(datatype)){
```
hmottestad: Should check if comparing what happens when comparing just time with date.

PR #160, opened 2019-01-28
eclipse/rdf4j#1268 add config for setting behavior when target is undefined

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1268 .

* add a config option for handling undefined target
* 
*


hmottestad:
@jeenbroekstra would also like this as part of the 2.5 release. Looks like my imagination got the better of me, so ended up implementing something that wasn't according to the spec. Made it configurable.

Note to self, need to update the documentation.

PR #162, opened 2019-01-29
Fix for returning milliseconds in SPARQL SECONDS()

barthanssens:
This PR addresses GitHub issue: eclipse/rdf4j#1267 .

* Return milliseconds (and even more detailed fractions of a second) if set

### DIFF
--- a/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/datetime/Seconds.java
+++ b/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/function/datetime/Seconds.java
@@ -7,6 +7,7 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.query.algebra.evaluation.function.datetime;
 
+import java.math.BigDecimal;
 import javax.xml.datatype.DatatypeConstants;
 import javax.xml.datatype.XMLGregorianCalendar;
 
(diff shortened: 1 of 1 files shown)


barthanssens:
Any suggestions where to add tests ? 

Logically, it should go to the testsuite, more specifically the 
rdf4j-testsuite/sparql/src/main/resources/testcases-sparql-1.1-w3c/functions/data.ttl and seconds-01.srx, but this (testing of fractions of seconds) is not part of the W3C official test suite... (maybe it should)

So I suppose I better add this in a separate file/package in the testsuite...

hmottestad:
Dunno where to put the tests.

But looks good. Nice catch.

jeenbroekstra:
@barthanssens you could add it as a testcase to `ComplexSPARQLQueryTest`. Not the best name for test suite, but that's where I've typically put regression test cases not covered by the official W3C testsuite. 

FWIW this is also one of those things where I want to do some cleanup in how the test framework is setup.

jeenbroekstra:
As a short-term solution, also fine with me if you don't modify ComplexSPARQLQueryTest (which is in the testsuite repo) and instead just add a test case to `MemoryComplexSPARQLQueryTest` (in sparql-compliance, in the storage repo - so it can be part of this PR).

barthanssens:
thanks, I'll add a test

PR #163, opened 2019-02-01
Issues/1273 shacl benchmark

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1273 .

* made benchmarks run again
* minor adjustments for performance
* only persist memory store used for storing shacl rules if the base sail is persisted

### DIFF
--- a/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -244,7 +244,7 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 	}
 
-	synchronized void refreshShapes(SailRepositoryConnection shapesRepoConnection) throws SailException {
+	synchronized List<NodeShape> refreshShapes(SailRepositoryConnection shapesRepoConnection) throws SailException {
 		try (SailRepositoryConnection beforeCommitConnection = shapesRepo.getConnection()) {
 			long size = beforeCommitConnection.size();
 			if (size > 0) {
@@ -257,6 +257,7 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 		runInferencingSparqlQueries(shapesRepoConnection);
 		nodeShapes = NodeShape.Factory.getShapes(shapesRepoConnection, this.config);
+		return nodeShapes;
 	}
 
 	@Override
--- a/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -9,6 +9,7 @@
 package org.eclipse.rdf4j.sail.shacl;
 
 import org.apache.commons.io.IOUtils;
+import org.eclipse.rdf4j.IsolationLevels;
 import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.vocabulary.RDF4J;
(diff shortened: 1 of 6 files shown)


shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -217,42 +217,37 @@ public void initialize() throws SailException {
...
-			}
-			catch (IOException e) {
-				throw new SailConfigException(e);
+			String path = super.getBaseSail().getDataDir().getPath();
+			if (path.endsWith("/")) {
+				path = path.substring(0, path.length() - 1);
 			}
+			path = path + "-shapes-graph/";
+
+			shapesRepo = new SailRepository(new MemoryStore(new File(path)));
```
hmottestad: @jeenbroekstra I changed the code you wrote for persisting the shacl rules memory store. Not it only persists if the base sail persists. It's for two reasons.
 - performance
 - abide by the users intended use and not force writing to disk

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -337,25 +308,47 @@ public void prepare() throws SailException {
...
+		refreshShapes(shapesConnection);
 
-			if (!sail.isIgnoreNoShapesLoadedException()
-				&& ((!addedStatementsSet.isEmpty() || !removedStatementsSet.isEmpty())
-				&& sail.getNodeShapes().isEmpty())) {
-				throw new NoShapesLoadedException();
-			}
+		if (!sail.isIgnoreNoShapesLoadedException()
+			&& ((!addedStatementsSet.isEmpty() || !removedStatementsSet.isEmpty())
+			&& sail.getNodeShapes().isEmpty())) {
```
jeenbroekstra: Aren't you introducing a possible race condition here by removing the synchronized block around this? What if a parallel connection also modified the shapes data?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -217,42 +217,37 @@ public void initialize() throws SailException {
...
 		if (super.getBaseSail().getDataDir() != null) {
-			path = super.getBaseSail().getDataDir().getPath();
-		}
-		else {
-			try {
-				path = Files.createTempDirectory("shacl-shapes").toString();
-			}
-			catch (IOException e) {
-				throw new SailConfigException(e);
+			String path = super.getBaseSail().getDataDir().getPath();
```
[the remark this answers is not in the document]
hmottestad: It's a bit hard to read in the diff. Code looks like this:

```java
if (super.getBaseSail().getDataDir() != null) {
	String path = super.getBaseSail().getDataDir().getPath();
	if (path.endsWith("/")) {
		path = path.substring(0, path.length() - 1);
	}
	path = path + "-shapes-graph/";

	shapesRepo = new SailRepository(new MemoryStore(new File(path)));
} else {
	shapesRepo = new SailRepository(new MemoryStore());
}
```

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
hmottestad: There is a lot of issues with race conditions with the shapes being cached in the ShaclSail and not in the connection. I'll make a new issue to track the work to for that.

PR #165, opened 2019-02-03
eclipse/rdf4j#1277 unify ShaclSailConfig and enable RDF-based config

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#1277 .

* unified two ShaclSailConfig classes. 
* ShaclSail remembers its own settings internally (instead of via a reference to a config object)
* enabled configuration via RDF-based configuration data (necessary for use by RepositoryManager, and in Server / Console)


shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -25,113 +31,96 @@
...
+ * 							&quot;@prefix xsd: &lt;http://www.w3.org/2001/XMLSchema#&gt; .&quot;,
+ * 							&quot;@prefix foaf: &lt;http://xmlns.com/foaf/0.1/&gt;.&quot;,
  *
- *					&quot;ex:PersonShapeProperty &quot;,
- *					&quot;	sh:path foaf:age ;&quot;,
- *					&quot;	sh:datatype xsd:int ;&quot;,
- *					&quot;	sh:maxCount 1 ;&quot;,
- *					&quot;	sh:minCount 1 .&quot;
- *				));
+ * 							&quot;ex:PersonShape&quot;, &quot;	a sh:NodeShape  ;&quot;, &quot;	sh:targetClass foaf:Person ;&quot;,
```
hmottestad: I'm wondering if you made any changes or intended to reformat this example code. Looks like maybe your automatic code reformatter had some fun here.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -337,76 +346,103 @@ private void runInferencingSparqlQueries(SailRepositoryConnection shaclSailConne
 	}
 
 	/**
-	 * 	Log (INFO) every execution step of the SHACL validation
-	 * 	This is fairly costly and should not be used in production.
-	 * Recommended to disable parallel validation with setParallelValidation(false)
+	 * Log (INFO) every execution step of the SHACL validation This is fairly costly and should not be used in
```
hmottestad: looks like this comment got a bit mangled too.

hmottestad:
This is fine by me. Thanks for cleaning up and unifying this @jeenbroekstra. It does look like your code formatter is a bit aggressive when it comes to the javadocs. I commented two places, but it looks like there are more.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra: Grmbl. It shouldn't touch javadoc code examples inside pre tags. I'll fix this.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra: That is actually as intended. Javadoc, like the code itself, has a line width set at 110 chars. If you want a hard line-break, you need to use a p tag or a br. In this case, I'll add the missing full stop :)

jeenbroekstra:
Thanks for reviewing @hmottestad, I think I took care of the most egregious formatting problems, and also found a minor bug in the handling of defaults.

hmottestad:
I reckon you can merge this.

PR #167, opened 2019-02-07
Sync develop with master

barthanssens:
It's been a while, so perhaps a good idea to merge the changes from master into develop


hmottestad:
go for it - as far as i know develop is broken due to some circular dep

PR #169, opened 2019-02-10
Issues/1286 1287 optimize parsing and skip validation

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1286  eclipse/rdf4j#1287.

* parse shacl predicates into a java object
* better support for requiresEvaluation
* skip validation if there are no changes


hmottestad:
`requiresEvaluation` is used to filter out shapes that don't need to be executed. Eg. a shape that validates `foaf:Person min 1 foaf:age` does not need to be executed when adding `:eclipse a foaf:Organization`

PR #171, opened 2019-02-12
Fix eclipse/rdf4j#1293 optimizer for regex calls now handles non-var arguments

JervenBolleman:
This PR addresses GitHub issue: eclipse/rdf4j#1293 .

The previous optimizer for replacing regex calls with contains,strstarts or strends did not work if the argument to the Regex was not a var.
e.g. REGEX(STR(?x), 'lala') 
was not optimized.
This fixes this and adds a test for it.

### DIFF
--- a/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizer.java
+++ b/evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizer.java
@@ -51,7 +51,7 @@ public class RegexAsStringFunctionOptimizer implements QueryOptimizer {
             final ValueExpr flagsArg = node.getFlagsArg();
             if (flagsArg == null || flagsArg.toString().isEmpty()) {
                 //if we have no flags then we can not be in case insensitive mode
-                if (node.getPatternArg() instanceof ValueConstant && node.getArg() instanceof Var) {
+                if (node.getPatternArg() instanceof ValueConstant) {
                     ValueConstant vc = (ValueConstant) node.getPatternArg();
                     String regex = vc.getValue().stringValue();
                     final boolean anchoredAtStart = regex.startsWith("^");
--- a/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizierTest.java
+++ b/evaluation/src/test/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizierTest.java
@@ -88,6 +88,16 @@ public class RegexAsStringFunctionOptimizierTest {
 
         testOptimizer(unoptimizedQuery, unoptimizedQuery);
     }
+    
+    @Test
+    public void testContainsFunction()
+            throws MalformedQueryException {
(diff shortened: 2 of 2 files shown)


JervenBolleman:
I am sorry for all the force pushes I am trying to get IP-validation part correct.

hmottestad:
good to see you got the signing issues fixed :) I

evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizer.java
```diff
@@ -51,7 +51,7 @@ public void meet(Regex node) {
             final ValueExpr flagsArg = node.getFlagsArg();
             if (flagsArg == null || flagsArg.toString().isEmpty()) {
                 //if we have no flags then we can not be in case insensitive mode
-                if (node.getPatternArg() instanceof ValueConstant && node.getArg() instanceof Var) {
```
hmottestad: Someone with a bit more knowledge needs to figure out what the intention for that last instanceof check was. Wrt it being same to remove now. If anyone knows @jeenbroekstra will know.

hmottestad: Ahh. I see that @JervenBolleman is the author of the class already. So I reckon this change should be perfectly fine.

hmottestad:
Should we consider targeting this at master and push it out with the 2.5 release?

hmottestad:
Also. What happens to Jenkins? Shouldn’t it be posting here asking if it needs to build and test this PR?

evaluation/src/main/java/org/eclipse/rdf4j/query/algebra/evaluation/impl/RegexAsStringFunctionOptimizer.java
JervenBolleman: The original Var check was from our code frontend at sparql.uniprot.org which runs with virtuoso and I need to change a sparql query string. Here I have an lovely AST and that is not required. So it was part of our own inline code for string mangeling.

jeenbroekstra:
@hmottestad (on my phone so forgive brevity) I'm not sure why it's not doing that, but you or Bart should be able to ping it by commenting '@rdf4j-bot run tests'

hmottestad:
@rdf4j-bot run tests

jeenbroekstra:
On a conference but I have a few spare minutes right now, I'll take a look at jenkins and why it's not picking up this PR

jeenbroekstra:
@rdf4j-bot run tests

jeenbroekstra:
Right, looks like I forgot to add you as an admin for the Jenkins PR config for rdf4j-storage, @hmottestad - that's why it ignored you. I've added you so the magic phrase should work for you too, from now on. FWIW Jenkins only automatically builds PRs from users that have been whitelisted, and Jerven hasn't yet.

jeenbroekstra:
Unfortunately this can not be easily rebased to the master branch as it was implemented off of a post 3.0 version of the develop branch. If we want to include it in 2.5, we'll need to cherrypick.

hmottestad:
Merged into develop and cherry picked into master.

PR #175, opened 2019-02-19
eclipse/rdf4j#1298 rdfs reasoner for the SHACL engine

hmottestad:
SHACL specs require that the engine implement a rudimentary RDFS reasoner that supports rdfs:subClassOf.

This PR addresses GitHub issue: eclipse/rdf4j#1298 .

* simple rdfs reasoner
* forward-chaining the Added and Removed memory stores
* backwards chain the base sail (hasStatement and getStatements)
* backwards chain SPARQL queries by using UNION in the SHACL implementation instead of attempting to do query rewrite after-the-fact


hmottestad:
Don't have any benchmarks for this, and it's not really been optimised for performance.

hmottestad:
Benchmarks show that unless you are adding `rdfs:subClassOf` statements there is no impact on performance.

PR #180, opened 2019-03-02
Support for NativeStore

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1314 .

* don't throw exeception during init() for reloading shapes
* run validation in serial on NativeStore
* tests

### DIFF
--- a/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -115,7 +115,7 @@ import java.util.List;
  *			connection.add(invalidSampleData, &quot;&quot;, RDFFormat.TURTLE);
  *			try {
  *				connection.commit();
- *			} catch (RepositoryException exception) {
+ *            } catch (RepositoryException exception) {
  *				Throwable cause = exception.getCause();
  *				if (cause instanceof ShaclSailValidationException) {
  *					ValidationReport validationReport = ((ShaclSailValidationException) cause).getValidationReport();
@@ -123,11 +123,11 @@ import java.util.List;
  *					// use validationReport or validationReportModel to understand validation violations
  *
  *					Rio.write(validationReportModel, System.out, RDFFormat.TURTLE);
- *				}
+ *                }
  *				throw exception;
- *			}
- *		}
- *	}
+ *            }
+ *        }
+ *    }
  *}
  *</pre>
  *
@@ -168,29 +168,22 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 	private boolean rdfsSubClassReasoning = ShaclSailConfig.RDFS_SUB_CLASS_REASONING_DEFAULT;
 
+	private boolean initializing = false;
+
 	static {
 		try {
-			SH_OR_UPDATE_QUERY = IOUtils.toString(
-					ShaclSail.class.getClassLoader().getResourceAsStream("shacl-sparql-inference/sh_or.rq"),
-					"UTF-8");
-			SH_OR_NODE_SHAPE_UPDATE_QUERY = IOUtils.toString(
-					ShaclSail.class.getClassLoader().getResourceAsStream(
-							"shacl-sparql-inference/sh_or_node_shape.rq"),
-					"UTF-8");
-			IMPLICIT_TARGET_CLASS_NODE_SHAPE = IOUtils.toString(
-					ShaclSail.class.getClassLoader().getResourceAsStream(
-							"shacl-sparql-inference/implicitTargetClassNodeShape.rq"),
-					"UTF-8");
-
-			IMPLICIT_TARGET_CLASS_PROPERTY_SHAPE = IOUtils.toString(
-					ShaclSail.class.getClassLoader().getResourceAsStream(
-							"shacl-sparql-inference/implicitTargetClassPropertyShape.rq"),
-					"UTF-8");
-		}
-		catch (IOException e) {
+			SH_OR_UPDATE_QUERY = resourceAsString("shacl-sparql-inference/sh_or.rq");
+			SH_OR_NODE_SHAPE_UPDATE_QUERY = resourceAsString("shacl-sparql-inference/sh_or_node_shape.rq");
+			IMPLICIT_TARGET_CLASS_NODE_SHAPE = resourceAsString("shacl-sparql-inference/implicitTargetClassNodeShape.rq");
+			IMPLICIT_TARGET_CLASS_PROPERTY_SHAPE = resourceAsString("shacl-sparql-inference/implicitTargetClassPropertyShape.rq");
+		} catch (IOException e) {
 			throw new IllegalStateException(e);
 		}
 
+	}
+
+	private static String resourceAsString(String s) throws IOException {
+		return IOUtils.toString(ShaclSail.class.getClassLoader().getResourceAsStream(s), "UTF-8");
 	}
 
 	public ShaclSail(NotifyingSail baseSail) {
@@ -205,7 +198,7 @@ public class ShaclSail extends NotifyingSailWrapper {
 	/**
 	 * Lists the predicates that have been implemented in the ShaclSail. All of these, and all combinations,
 	 * <i>should</i> work, please report any bugs. For sh:path, only single predicate paths are supported.
-	 * 
+	 *
 	 * @return List of IRIs (SHACL predicates)
 	 */
 	public static List<IRI> getSupportedShaclPredicates() {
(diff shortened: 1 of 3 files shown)


hmottestad:
@jeenbroekstra I've found out that connections will often fail if they are used for read operations by multiple threads. SHACL validation in parallel uses multiple threads against the same connection, I'm not sure exactly how to do parallel validation without sharing a connection. Any suggestions?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -249,32 +252,33 @@ public void initialize() throws SailException {
...
-	{
-		try (SailRepositoryConnection beforeCommitConnection = shapesRepo.getConnection()) {
-			long size = beforeCommitConnection.size();
-			if (size > 0) {
-				// Our inferencer both adds and removes statements.
-				// To support updates I recommend having two graphs, one raw one with the unmodified data.
-				// Then copy all that data into a new graph, run inferencing on that graph and use it to generate the java objects
-				throw new IllegalStateException(
+		throws SailException {
+		if (!initializing) {
```
hmottestad: this is the key change

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -234,8 +227,18 @@ public ShaclSail() {
 
 	@Override
 	public void initialize() throws SailException {
+		init();
+	}
+
+	@Override
+	public void init() throws SailException {
+		initializing = true;
```
jeenbroekstra: Couple of things:

1. you shouldn't override `init()` directly. Instead, override`initializeInternal()`.
2. you don't need a new boolean flag here: you can just the `isInitialized()` protected boolean flag inherited from `AbstractSail`. Additional advantage is that `AbstractSail` ensures that reading that flag is thread-safe (if you don't override `init()` that is :) ).

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -234,8 +227,18 @@ public ShaclSail() {
 
 	@Override
 	public void initialize() throws SailException {
+		init();
```
jeenbroekstra: This override shouldn't be necessary - `AbstractSail` does this.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra:
```suggestion
		if (!this.isInitialized()) {
```

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/NativeStoreTest.java
```diff
@@ -0,0 +1,124 @@
...
+		SailRepository shaclSail = new SailRepository(new ShaclSail(new NativeStore(file)));
+		shaclSail.init();
+
+		shaclSail.shutDown();
+
+		delete(file);
+	}
+
+	@Test(expected = ShaclSailValidationException.class)
+	public void testPresistedShapes() throws Throwable {
```
jeenbroekstra:
```suggestion
	public void testPersistedShapes() throws Throwable {
```

jeenbroekstra: The problem doesn't seem specific to the native store, but to _any_ store that persists data / has a datadir. I don't have any objection to using the native store as the test case, but wouldn't you have the same issue when using a memory store with persistence enabled?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -249,32 +252,33 @@ public void initialize() throws SailException {
...
-				throw new IllegalStateException(
+		throws SailException {
+		if (!initializing) {
+			try (SailRepositoryConnection beforeCommitConnection = shapesRepo.getConnection()) {
+				boolean empty = !beforeCommitConnection.hasStatement(null, null, null, false);
+				if (!empty) {
+					// Our inferencer both adds and removes statements.
+					// To support updates I recommend having two graphs, one raw one with the unmodified data.
+					// Then copy all that data into a new graph, run inferencing on that graph and use it to generate the java objects
+					throw new IllegalStateException(
```
jeenbroekstra: This is a bit beyond the scope of this current fix perhaps, but couldn't we just reset internal state when shapes get modified so that we know we need to do a full data validation "refresh", instead of just throwing an IllegalStateException?

jeenbroekstra:
> SHACL validation in parallel uses multiple threads against the same connection, I'm not sure exactly how to do parallel validation without sharing a connection.

Generally speaking, connections are not thread-safe. If you wish to share one for concurrent operations, you will have to put sufficient guards and locking in at your end. But the whole thing is a bit of a code smell to be honest. Why share the same connection in multiple threads at all? If the goal is to be able to "see" modifications across threads before they are committed, just open a connection with the right isolation level (`READ_UNCOMMITTED`). This kind of use is exactly what those isolation levels were designed for :)

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
hmottestad: There is no support for revalidating the entire underlying store yet. Implementing support for sh:or helped, so it'll be easier to implement in the future.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
hmottestad: does not seem to be a isInitialized() method

hmottestad: There isn't any AbstractSail in the inheritance hierarchy either.

hmottestad: No initializeInternal() either. I guess all that stuff comes from the AbstractSail, but I'm extending NotifyingSailWrapper.

jeenbroekstra: Well, crap (excuse my Klatchian), that is the first time I've noticed that SailWrapper does not extend AbstractSail. Apologies. 

Ok. In that case stick with this for now. You may want to make the `private boolean initializing` field `volatile` and make `init()` and  `refreshShapes()` use a common mutex to avoid a race condition on the state of the flag.

hmottestad:
> Generally speaking, connections are not thread-safe. If you wish to share one for concurrent operations, you will have to put sufficient guards and locking in at your end. But the whole thing is a bit of a code smell to be honest. Why share the same connection in multiple threads at all? If the goal is to be able to "see" modifications across threads before they are committed, just open a connection with the right isolation level (`READ_UNCOMMITTED`). This kind of use is exactly what those isolation levels were designed for :)

I'm not sure using multiple connections will work very easily. And as far as I remember ´READ_UNCOMMITTED` means "might read uncommitted" and not "will read uncommitted"; this was what I discovered on my endeavour to create a new memory store. And ´READ_UNCOMMITTED` will also read from other transactions, which isn't necessarily what the user expects. 

For the ShaclSail, the user opens a transaction and makes their changes and then calls `commit()`. The ShaclSail then generates and runs multiple query plans for validating those changes. The queries need to run against the current transaction, and to be completely safe that transaction should actually be `SERIALIZABLE` (especially for things like maxCount). Running said queries in parallel is the simplest way to speed up the validation.  

How much would need to be `synchronized` to make a connection thread safe? Can iterations run in parallel or do those also need to be `synchronized`?

hmottestad:
Btw. I tried applying your commits with the github web gui. Email and sign-off is broken so that eclipse bot complains.

jeenbroekstra:
@hmottestad too bad, we'd better not use that feature again then.

jeenbroekstra:
I'm not sure it can be done in a way that still makes parallel validation worthwhile to be honest.

hmottestad:
I’ve had a little look at synchronisation, but couldn’t get it to work. The deadlock in the code looks like it is applicable to both the native and the MemoryStore. So I think maybe we should disable parallel shacl in general for now (I’ll call it an experimental feature). 

The deadlock in the code might be worth looking into though. Looks like it’s a classic thread1 grabs lockA and thread2 grabs lockB, then they try to grab each other’s locks. I think it’s in SailSource. Might be possible to fix with ordering of locks. 

I’ll merge this branch now, and then make a second one where I make parallel default false.

PR #181, opened 2019-03-03
SHACL parallel validation false

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1315 .

* make default false
* log warning and info


hmottestad:
What do you think @jeenbroekstra ?

jeenbroekstra:
Code looks ok at first glance but all tests fail with a Stackoverflow, I think this introduced an infinite regression somewhere?

PR #183, opened 2019-03-06
Issues/1319 shacl deadlock

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1319 .

* make a deadlock test
* move super.prepare() and previousStateConnection.prepare() to final step in prepare()

### DIFF
--- a/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
+++ b/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
@@ -373,32 +373,34 @@ public class ShaclSailConnection extends NotifyingSailConnectionWrapper implemen
 
 	@Override
 	public void prepare() throws SailException {
-		preparedHasRun = true;
-		super.prepare();
-		previousStateConnection.prepare();
+		try {
+			preparedHasRun = true;
 
-		List<NodeShape> nodeShapes = refreshShapes(shapesConnection);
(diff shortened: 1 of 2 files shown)


hmottestad:
@jeenbroekstra I managed to find out what was causing the deadlock. At some point down the line of prepare() there is a semaphore.lock() without a corresponding unlock(). There is an unlock(), but I think it's called from somewhere else, as part of a close() process. Every once in a while we would then get a deadlock (and never when debugging). 

Solution was to first validate all the date and then call super.prepare().

I'm not sure this fixes all cases of deadlock, but I haven't been able to produce any others.

Are you comfortable with pushing this into the 2.5 release, or do you want to wait for a 2.5.x or 3.0?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -373,33 +373,35 @@ synchronized public void close() throws SailException {
...
+			boolean valid = invalidTuples.isEmpty();
 
-		if (!valid) {
-			throw new ShaclSailValidationException(invalidTuples);
+			if (!valid) {
+				throw new ShaclSailValidationException(invalidTuples);
+			}
+		}finally {
+			super.prepare();
+			previousStateConnection.prepare();
```
hmottestad:
```
super.prepare();
previousStateConnection.prepare();
```

As the final step in `prepare()`

jeenbroekstra:
Looks good, great find! Let's merge this in.

PR #192, opened 2019-03-10
Issues/1337 faster spin

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1337 .

Happy leet :)

* Spin is now 1.8x faster

It's still stupid slow though. Takes 130ms to init(). MemoryStore takes 0.02ms!


hmottestad:
@JervenBolleman @dwvisser @mhale @jeenbroekstra 

I've been working on optimizing the SpinSail. It's seriously slow, and I've made it a bit less slow. It's still, as I would say, stupidly slow.

It seems like you 4 have worked on the SpinSail, so I have a couple of questions for you. I would really appreciate if you had the time to help me out :)

 - Is it OK to skip checkConstraints() during initialization of the Sail? Is it OK even when the base sail is not empty? Will modifications after initialization check all constraints against those changes or maybe all the data?

 - On initialize() the SpinSail will load some base files. It does inference and checkConstraints() against those changes. Would it be OK to simply forward-chain these files and load that forward-chained result for quicker initialization? Basically, will changes to the sail load required functions (and the like) when needed, or does all the spin stuff get loaded when the actual spin files are loaded?

sail-spin/src/main/java/org/eclipse/rdf4j/sail/spin/SpinSail.java
```diff
@@ -45,6 +45,8 @@
 
 	private boolean axiomClosureNeeded = true;
 
+	private boolean initializing;
```
hmottestad: Make volatile.

jeenbroekstra:
Fwiw I have done very little work on this, most of it was developed by Mark Hale I think. From the top of my head there shouldn't be an issue with removing verification on init().

jeenbroekstra:
@rdf4j-bot run tests

hmottestad:
looks like a bit of whack-a-mole with the reasoner :(

hmottestad:
Current benchmark status on develop:

```
# Run complete. Total time: 00:05:54

Benchmark                                   Mode  Cnt      Score     Error  Units
BasicBenchmarks.addRemove                   avgt    5  11755.278 ± 166.660  ms/op
BasicBenchmarks.addRemoveAll                avgt    5   1487.938 ±  29.380  ms/op
BasicBenchmarks.addRemoveSingleTransaction  avgt    5    772.273 ±  24.008  ms/op
BasicBenchmarks.addRemoveTwoTransactions    avgt    5   1592.911 ±  50.987  ms/op
BasicBenchmarks.cantInvoke                  avgt    5   7880.310 ± 173.010  ms/op
BasicBenchmarks.cantInvokeOldReasoner       avgt    5   2819.196 ± 156.742  ms/op
BasicBenchmarks.memoryStoreInit             avgt    5      0.031 ±   0.042  ms/op
BasicBenchmarks.remove                      avgt    5    782.293 ±  44.910  ms/op
BasicBenchmarks.spinSailInit                avgt    5    739.388 ±  29.887  ms/op
```

hmottestad:
Current state of this branch.

```
# Run complete. Total time: 00:03:33

Benchmark                                   Mode  Cnt     Score     Error  Units
BasicBenchmarks.addRemove                   avgt    5  8457.807 ± 267.835  ms/op
BasicBenchmarks.addRemoveAll                avgt    5   478.263 ±  31.521  ms/op
BasicBenchmarks.addRemoveSingleTransaction  avgt    5    54.495 ±   4.489  ms/op
BasicBenchmarks.addRemoveTwoTransactions    avgt    5   599.575 ±  31.950  ms/op
BasicBenchmarks.cantInvoke                  avgt    5  1099.050 ± 180.946  ms/op
BasicBenchmarks.cantInvokeOldReasoner       avgt    5  2142.362 ±  56.377  ms/op
BasicBenchmarks.memoryStoreInit             avgt    5     0.031 ±   0.038  ms/op
BasicBenchmarks.remove                      avgt    5    50.258 ±   3.939  ms/op
BasicBenchmarks.spinSailInit                avgt    5    52.167 ±   2.373  ms/op
```

hmottestad:
Why this one is so slow is a bit of a mystery:

```
BasicBenchmarks.addRemove                   avgt    5  8457.807 ± 267.835  ms/op
```

hmottestad:
Screenshot 2019-03-12 at 16 15 56

Seems to be an issue with cleaning snapshots.

hmottestad:
@jeenbroekstra Ok. So here is where I stand with the reasoner right now. Fixing it so it works with the SPIN sail has unearthed a race condition in the way rollback is handled, and fixing that causes a transaction-not-started error. I'm sorry I didn't find this before we made the reasoner the default one, but I've got a good grasp on it now so I think fixing it will be very possible. Just going to need a few days. After that I will port it to master for the next patch release (2.5.1).

jeenbroekstra:
Even when done, can you hold off on merging until _after_ the code formatting stuff has been applied to both master and develop? I want to avoid us getting into merge conflict hell.

sail-base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
```diff
@@ -235,7 +235,10 @@ public synchronized void clear(Resource... contexts) {
 	@Override
 	public synchronized void approve(Resource subj, IRI pred, Value obj, Resource ctx) {
 		if (deprecated != null) {
-			deprecated.remove(subj, pred, obj, ctx);
+			boolean readdedARemovedStatement = deprecated.remove(subj, pred, obj, ctx);
+			if (readdedARemovedStatement) {
+				return;
+			}
```
hmottestad: @jeenbroekstra this here helps reduce the thrashing in the spinsail a lot. Problem is that ever time something is removed it will remove all 10 000 or so preloaded spin rules and then add then in again. Essentially because it doesn’t keep track or where an inferred triple was generated from, so it will have to clear all inferred triples every time something is removed. 

Is this an ok change to make?

hmottestad:
@rdf4j-bot run tests

sail-base/src/main/java/org/eclipse/rdf4j/sail/base/Changeset.java
hmottestad: Ive tested this a bit and found out that chnageset is not the correct place to handle this. The problem boils down to that if you remove a nonexistent triple and then add it again, it won’t be added!

So I need to keep digging.

hmottestad:
Currently not managing to get much further on this. Might just call it quits for now. Just too many layers to peel off to find out exactly how the inferred statements are added and removed.

hmottestad:
@rdf4j-bot run tests

PR #194, opened 2019-03-13
Formatting settings

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#822 and eclipse/rdf4j#1240 .

* added maven formatter plugin
* reformatter all code
* updated readme and contributing guidelines


shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -40,98 +40,88 @@
...
- *	}
- *}
- *</pre>
+ * 					Rio.write(validationReportModel, System.out, RDFFormat.TURTLE);
+ * 				}
+ * 				throw exception;
+ * 			}
+ * 		}
+ * 	}
+ * }
```
hmottestad: Can you double check how this ends up looking in the javadocs?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -196,27 +188,10 @@ public ShaclSail() {
...
-			SHACL.MIN_INCLUSIVE,
-			SHACL.MAX_EXCLUSIVE,
-			SHACL.MAX_INCLUSIVE,
-			SHACL.CLASS,
-			SHACL.TARGET_NODE
-		);
+		return Arrays.asList(SHACL.TARGET_CLASS, SHACL.PATH, SHACL.PROPERTY, SHACL.OR, SHACL.MIN_COUNT, SHACL.MAX_COUNT,
+				SHACL.MIN_LENGTH, SHACL.MAX_LENGTH, SHACL.PATTERN, SHACL.FLAGS, SHACL.NODE_KIND_PROP, SHACL.LANGUAGE_IN,
+				SHACL.DATATYPE, SHACL.MIN_EXCLUSIVE, SHACL.MIN_INCLUSIVE, SHACL.MAX_EXCLUSIVE, SHACL.MAX_INCLUSIVE,
+				SHACL.CLASS, SHACL.TARGET_NODE);
```
hmottestad: Can you revert this one? It’s a lot easer to read and change when it’s a vertical list.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -225,8 +200,9 @@ public void initialize() throws SailException {
 		super.initialize();
 
 		if (getDataDir() != null) {
-			if(parallelValidation){
-				logger.info("Automatically disabled parallel SHACL validation because persistent base sail was detected! Re-enable by calling setParallelValidation(true) after calling init() / initialize().");
+			if (parallelValidation) {
+				logger.info(
+						"Automatically disabled parallel SHACL validation because persistent base sail was detected! Re-enable by calling setParallelValidation(true) after calling init() / initialize().");
```
hmottestad: So it doesn’t handle long strings very well.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -262,49 +259,46 @@ void cleanup() {
...
 
 			try {
-				Stream<PlanNode> planNodeStream = sail
-					.getNodeShapes()
-					.stream()
-					.flatMap(nodeShape -> nodeShape.generatePlans(this, nodeShape, sail.isLogValidationPlans()).stream());
+				Stream<PlanNode> planNodeStream = sail.getNodeShapes()
+						.stream()
+						.flatMap(nodeShape -> nodeShape.generatePlans(this, nodeShape, sail.isLogValidationPlans())
+								.stream());
```
hmottestad: That last line break there is quite annoying.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -337,21 +331,23 @@ void fillAddedAndRemovedStatementRepositories() {
...
-				.stream()
-				.filter(statement -> !removedStatementsSet.contains(statement))
-				.flatMap(statement -> rdfsSubClassOfReasoner == null ? Stream.of(statement) : rdfsSubClassOfReasoner.forwardChain(statement))
-				.forEach(statement -> connection.addStatement(statement.getSubject(), statement.getPredicate(), statement.getObject(), statement.getContext()));
+			addedStatementsSet.stream()
+					.filter(statement -> !removedStatementsSet.contains(statement))
+					.flatMap(statement -> rdfsSubClassOfReasoner == null ? Stream.of(statement)
+							: rdfsSubClassOfReasoner.forwardChain(statement))
+					.forEach(statement -> connection.addStatement(statement.getSubject(), statement.getPredicate(),
+							statement.getObject(), statement.getContext()));
```
hmottestad: Quite a few of these that line break in rather ugly ways.

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/AbstractShaclTest.java
```diff
@@ -38,53 +38,30 @@
...
+			"test-cases/or/inheritance-deep", "test-cases/or/inheritance-deep-minCountMaxCount",
+			"test-cases/or/inheritanceNodeShape", "test-cases/or/datatype", "test-cases/or/datatypeTargetNode",
+			"test-cases/or/minCountMaxCount", "test-cases/or/maxCount", "test-cases/or/minCount",
+			"test-cases/or/nodeKindMinLength", "test-cases/or/implicitAnd", "test-cases/or/datatypeDifferentPaths",
+			"test-cases/minExclusive/simple", "test-cases/minExclusive/dateVsTime", "test-cases/maxExclusive/simple",
+			"test-cases/minInclusive/simple", "test-cases/maxInclusive/simple", "test-cases/implicitTargetClass/simple",
+			"test-cases/class/simple", "test-cases/class/subclass", "test-cases/class/targetNode",
+			"test-cases/or/class", "test-cases/or/datatype2", "test-cases/or/minCountDifferentPath",
+			"test-cases/datatype/targetNode"
+
```
hmottestad: My usecase here is to add a new test in the “correct” place and then comment out all the other ones while I work. There must be some way to force it to break long lists like this on every argument and not on line length. Will also make diffs in git easer to read.

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/AbstractShaclTest.java
```diff
@@ -292,8 +267,7 @@ String getShaclPath() {
 	}
 
 	enum ExpectedResult {
-		valid,
-		invalid
+		valid, invalid
```
hmottestad: Enums need to be on separate lines. That must be configurable.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
```diff
@@ -69,30 +70,37 @@ public boolean requiresEvaluation(SailConnection addedStatements, SailConnection
...
-		stringBuilder.append(System.identityHashCode(removedStatements) + " [label=\"Removed statements\" nodeShape=pentagon fillcolor=lightblue style=filled];").append("\n");
-
+		stringBuilder
+				.append(System.identityHashCode(addedStatements)
+						+ " [label=\"Added statements\" nodeShape=pentagon fillcolor=lightblue style=filled];")
+				.append("\n");
+		stringBuilder
+				.append(System.identityHashCode(removedStatements)
+						+ " [label=\"Removed statements\" nodeShape=pentagon fillcolor=lightblue style=filled];")
+				.append("\n");
```
hmottestad: I feel that this definitely makes it less readable. Wasn’t very readable to begin with. But worse now.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
```diff
@@ -131,76 +137,83 @@ public SourceConstraintComponent getSourceConstraintComponent() {
...
+					.stream(connection.getStatements(ShapeId, SHACL.PROPERTY, null))) {
+				return stream.map(Statement::getObject).map(v -> (Resource) v).flatMap(propertyShapeId -> {
+					List<PropertyShape> propertyShapes = getPropertyShapesInner(connection, nodeShape, propertyShapeId);
 
-						return propertyShapes.stream();
+					return propertyShapes.stream();
 
-					})
-					.collect(Collectors.toList());
+				}).collect(Collectors.toList());
```
hmottestad: This use of stream breaks incorrectly and becomes unnecessarily hard to read.

hmottestad: Github isn’t showing it here, but it’s the 6-8 lines above.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra: I'll have a look at this, this might be a glitch.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
jeenbroekstra: Agreed. I'm not quite sure where the linebreak between `stringBuilder` and `.append` comes from - that should remain a single line. I'll take a closer look.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra: I'll have a look at making this a general rule. I'll also take another look at reconfiguring the formatter to not join up two lines if they have been manually split - not sure I can make that work for all cases but I'll try.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
hmottestad: ‘’’
		static List<PropertyShape> getPropertyShapes(Resource ShapeId, SailRepositoryConnection connection,
				NodeShape nodeShape) {

			try (Stream<Statement> stream = Iterations
					.stream(connection.getStatements(ShapeId, SHACL.PROPERTY, null))) {
				return stream.map(Statement::getObject).map(v -> (Resource) v).flatMap(propertyShapeId -> {
					List<PropertyShape> propertyShapes = getPropertyShapesInner(connection, nodeShape, propertyShapeId);

					return propertyShapes.stream();

				}).collect(Collectors.toList());
			}

		}
‘’’

hmottestad: https://github.com/eclipse/rdf4j-storage/blob/57c7e098226e41006bdec12dd10d0b38d3884fad/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java#L140

hmottestad: This is how it used to look: 

```
static List<PropertyShape> getPropertyShapes(Resource ShapeId, SailRepositoryConnection connection, NodeShape nodeShape) {

	try (Stream<Statement> stream = Iterations.stream(connection.getStatements(ShapeId, SHACL.PROPERTY, null))) {
		return stream
			.map(Statement::getObject)
			.map(v -> (Resource) v)
			.flatMap(propertyShapeId -> {
				List<PropertyShape> propertyShapes = getPropertyShapesInner(connection, nodeShape, propertyShapeId);

				return propertyShapes.stream();

			})
			.collect(Collectors.toList());
	}

}
```

hmottestad: I'm looking at preserving the vertical listing of the calls to the second stream. The fact that it breaks in the `try` section because it is too long is fine. 

I think what I'm trying to grasp at is that I like it when I can read either mostly vertical or mostly horizontal when trying to understand what a piece of code does. So each "logical" block of code that I'm trying to understand should either be completely horizontal or completely vertical.  

I also like not having to read backwards to be able to understand something. So when the formatter chains all the .map .map .flatmap on a single line and then the body of the .flatmap on a new line, I now have to read the line backwards for parenthesis and fullstops to see where the method for the body starts.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra: There's no easy way around this I'm afraid. Long strings are a problem. In this case it's up to the developer to actually split the string.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
jeenbroekstra: Not something I can easily do something about in the formatting - the line is simply too long. 

However, I've tweaked the formatter so it will no longer join up lines if they are deliberately split by the developer. So you have options to split this differently yourself (for example, you could insert a newline after `nodeShape.generatePlans(this, ` and then join the `.stream()` back up with the rest of the line.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
jeenbroekstra: They're all too long. Needs developer intervention to re-write.

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/AbstractShaclTest.java
jeenbroekstra: This should be fixed now (as also with the array of constants in ShaclSail).

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/AbstractShaclTest.java
jeenbroekstra: Fixed in two ways: it will no longer join up separate lines, plus in the case of enums it will explicitly put each element on a separate line when formatting.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
jeenbroekstra: I had a closer look at this case and I'm afraid this again is something that only the developer can fix. The code is just intrinsically written without regard for a max line width (and also combines stringbuilder append with string concatenation which makes it less readable in general).

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/PropertyShape.java
jeenbroekstra: This should be fixed now, in the sense that the formatter will leave your linebreaks alone (merely adjusting indentation where necessary).

jeenbroekstra:
@hmottestad @barthanssens I have made the formatter less strict, and configured it to _not_ re-join two lines if the developer has explicitly split them. This means that as individual devs we have a lot more control over the details of formatting things like long lists/arrays, fluent/streaming api calls, etc.

compliance/store/src/test/java/org/eclipse/rdf4j/sail/memory/SchemaCachingRDFSInferencerMemInferencingTest.java
```diff
@@ -92,10 +92,9 @@ public void testBlankNodePredicateInference() {
 
 	@Test
 	public void testRollback()
-		throws NoSuchMethodException,
-		InvocationTargetException,
-		IllegalAccessException
-	{
+			throws NoSuchMethodException,
```
jeenbroekstra: Unfortunately due to the looser settings in the formatter, this kind of thing no longer automatically gets corrected. It will be up to developer discreation to put the throws clause _behind_ the method declaration instead of on the next line. Or not :)

hmottestad: That is a pity. We could always pivot RDF4J into a code reformatting framework, convert javacode intro triples and then use sparql to write it back out again as formatted code :P Just kidding. Maybe sometime in the future someone will make a truly awesome code formater that can handle all kinds of edge cases and has been tested on a large variety of code bases to format code for any and all situations. Or maybe even better, we go away from mixing the semantics of the code with the syntax, so that the semantics is stored in git and your IDE will show you the syntax based on your own preferences.

jeenbroekstra: One can dream :)

Actually what I'll do this weekend before merging this, is a one-time pass with the original, stricter formatting settings (at least the rejoining-lines bit), so that this gets corrected throughout the codebase - but the settings that will be applied on an ongoing basis will still be the looser ones. That way we get thing likes this checked in correctly. It may result in a few things being weirdly formatted this time around, but those will be the rare edge cases hopefully, and will be easier to correct by individual developers.

hmottestad: Ok

jeenbroekstra:
@rdf4j-bot run tests

PR #203, opened 2019-03-23
Issues/1290 revalidate added data

hmottestad:

* added a revalidate() method to allow the connection to be revalidated


hmottestad:
@jeenbroekstra @barthanssens 
This is the beginning of being able to validate an existing database, be able to validate using less memory, or be able to revalidate after changing the SHACL shapes.

The biggest current issue with this approach is that it runs SPARQL queries with "order by". And I'm not sure if the native store is smart enough to understand that it can use its sorted indexes for that instead of doing an explicit sort (which takes n log n time and n space).

Next steps will be to find out how to handle changes to the SHACL shapes.

PR #207, opened 2019-03-25
 Make LuceneSailConnection to replace query patterns with results inplace 

KMax:
This PR addresses GitHub issue: eclipse/rdf4j#1229 .

* Remove from LuceneSailConnection the code responsible for the inclusion of the query results into the query in form of BindingSetAssigment. Move this functionality to SearchQueryEvaluator subclasses.
* SearchQueryEvaluator subclasses replace the query patterns with the results *in place* instead of *"smart"* replacement mechanism used in LuceneSailConnection before.
* Add unit tests for the subclasses.


KMax:
Is there a way to run disabled tests mark with a comment "JTS is required"  like this [one](https://github.com/eclipse/rdf4j-storage/blob/master/compliance/elasticsearch/src/test/java/org/eclipse/rdf4j/sail/elasticsearch/ElasticsearchSailGeoSPARQLTest.java#L94)? It'd help in testing this PR.

hmottestad:
@rdf4j-bot run tests

jeenbroekstra:
@KMax those disabled tests are a leftover from when we couldn't ship JTS with RDF4J due to licensing issues. That problem has been resolved, JTS is now shipped with RDF4J. If you want to try just re-enabling those ignored tests (by removing the annotation) that should be fine.

KMax:
Re-enabled the tests. They fail with:

```
ERROR There was a problem evaluating spatial relation query 'http://www.opengis.net/def/function/geosparql/sfIntersects POLYGON ((2.315 48.855, 2.360 48.835, 2.370 48.850, 2.315 48.855))'!
java.text.ParseException: java.lang.UnsupportedOperationException: Unsupported shape of this SpatialContext. Try JTS or Geo3D.
```

Looking for a solution, but I can't find it yet.

hmottestad:
@rdf4j-bot run tests

jeenbroekstra:
@KMax sorry for the delay - it looks like one of your commits was not signed off, which makes Eclipse IP validation fail. Can you check and correct (easiest way to do that is to amend the problematic commit, and then force-push, I think)?

lucene-api/src/main/java/org/eclipse/rdf4j/sail/lucene/QuerySpec.java
```diff
@@ -9,19 +9,15 @@
 
 import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.Resource;
-import org.eclipse.rdf4j.query.algebra.EmptySet;
-import org.eclipse.rdf4j.query.algebra.QueryModelNode;
-import org.eclipse.rdf4j.query.algebra.SingletonSet;
-import org.eclipse.rdf4j.query.algebra.StatementPattern;
-import org.eclipse.rdf4j.query.algebra.Var;
+import org.eclipse.rdf4j.query.algebra.*;
```
jeenbroekstra: This is no big deal but we have a convention to not use wildcard imports. Happy to merge as-is but can you check your editor settings for future reference?

jeenbroekstra:
Apart from the ip validation, it looks like this greenlines. I'm happy to merge this as-is tbh, and treat re-enabling the ignored JTS tests as a separate issue.

hmottestad:
@rdf4j-bot run tests

jeenbroekstra:
We're having some conflicts I see. I'll take a closer look and try to merge and solve.

KMax:
@jeenbroekstra Thank you! I've merged it locally, but I faced with a failing test and couldn't fix it quickly, so I decided to do it later when I have more time.

jeenbroekstra:
I'm not sure what has happened here exactly, but your PR is suddenly over 1,200 files, and from a glance it looks as if it is trying to overwrite a lot of things with older versions (for example the contributors guide, and a couple of other things). I'll see if I can fix by cherry picking your changes.

jeenbroekstra:
@KMax it looks as if this is not something I can easily fix for you. For some reason, your commit 
0798a08, which is supposed to only contain the actual changes you mnade in the LuceneSail, now contains 1,299 changed files, all over the project. I think that something may have gone wrong when you force-pushed your amended commit. 

I must admit I'm not immediately sure how to fix this. I'm hoping that you have a way to easily retrieve the files that you _actually_ wanted to change. Since (from memory) it's only six or seven files, it might be easiest to just save those changes somewhere off to the side, then completely delete your branch and your PR, create a new one and manually commit your changes again, then put up a new PR. Sorry for the hassle. And I'm just guessing here, if you find a better way to fix it by all means do!

jeenbroekstra:
Fwiw if you're fed up and prefer to just send me a patch attachment so that I can commit the relevant fixes, that's also fine.

KMax:
@jeenbroekstra Yes, sorry for the mess with the commits, my fault. I'll fix it hopefully this week.

KMax:
Resubmitted PR, see https://github.com/eclipse/rdf4j-storage/pull/255

PR #211, opened 2019-03-26
Issues/1290 automatic revalidation

hmottestad:

* allow for modifying shapes after they are loaded
* revalidate all data if shapes are changed

Before:
```
Benchmark                                Mode  Cnt   Score   Error  Units
ComplexBenchmark.shaclEmptyTransactions  avgt   10  10.557 ± 0.260  ms/op
```

After:
```
Benchmark                                Mode  Cnt   Score   Error  Units
ComplexBenchmark.shaclEmptyTransactions  avgt   10  12.943 ± 0.318  ms/op
```


hmottestad:
@jeenbroekstra I could really use your eyes on this one. I hope that is thread safe. I realise it's not isolated very well, since a new transaction will read the current state of the shapes, even if the transaction that modified them ends up rolling back. I'll need to think about that one for a while. Caching is important for performance, but I don't want to cause any isolation issues because of it.

hmottestad:
I think I will take the approach we took with the RDFS reasoner: exclusive lock the moment the shapes graph is modified

I can also inline the “cache” memory store that is used to run various SPARQL update queries. I think I might also look into using RDFS for some  domain/range reasoning.

hmottestad:
@jeenbroekstra I've now implemented locking and isolation similarly as in the SchemaCachingRDFSInferencer. And made a couple of tests. I'm quite happy now.

hmottestad:
I have come to think of a edge case though. If you start two transactions T1 and T2. T1 modifies the shapes so that whatever T2 adds is invalid, but because T2 starts before T1 is finished then it sees the snapshot of shapes from before T1. 

T1 begin
T2 begin
T1 add shape
T1 commit
T2 add data (invalid according to new shapes in T1)
T2 commit
 No exceptions but sail is in invalid state.

hmottestad:
Maybe a readlock on prepare() and a write lock on any modifications to the shapes. That way all transactions would halt and wait for the transaction that modifies the shapes to finish. And then always get the cached shapes in prepare() instead of in begin().

jeenbroekstra:
I won't have time to really review until this weekend. But doing a global write lock on shape modifications sounds like a decent way forward.

hmottestad:
This is deadlock free in real world use cases: one connection per thread and no coordination between threads. 

If you have two connections in a thread you can deadlock the code:

T1 begin
T2 begin
T1 load shapes ( acquire writelock)
T2 add data
T2 commit ( deadlocks waiting for readlock)

We could add deadlock detection for this scenario. It would need to be based on detecting that there is a single thread with multiple connections. Simple time-based deadlock detection is not sufficient since the revalidation that runs when you modify shapes could take a long time!

hmottestad:
Needs cleanup of the stamp! And need to figure out what values for stamp are safe for default use.

hmottestad:
@jeenbroekstra I'll keep working on this and making some more tests and maybe some benchmarks. This is very important to get right, so I'll leave it open until you get a chance to review it.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -33,14 +35,16 @@
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.locks.ReentrantLock;
```
jeenbroekstra: I think this import is no longer used?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -292,78 +297,56 @@ public NotifyingSailConnection getConnection() throws SailException {
...
-		return this.logValidationPlans;
-	}
+	long acquireExclusiveWriteLock(long stamp) {
+		if (lock.validate(stamp)) {
+			return stamp;
+		}
 
-	public boolean isIgnoreNoShapesLoadedException() {
-		return this.ignoreNoShapesLoadedException;
+		return lock.writeLock();
```
jeenbroekstra: If you set a reference at this point to the thread obtaining the lock, you can avoid the deadlock edge case. E.g. something like this:

     synchronized { 
       if (Thread.currentThread().equals(this.threadHoldingLock)) {
          throw new SailConflictException("deadlock detected");
       } 
       long stamp = lock.writeLock();
       this.threadHoldingLock = Thread.currentThread();
       return stamp;
     }

hmottestad: I know handle the following two cases:
 - single thread tries to modify shapes in two overlapping connections/transactions
 - single thread tries to both modify shapes and commit data in two overlapping connections/transactions

I see that you added a synchronized block around the code, but I don't feel that this is necessary. The code doesn't care which thread is the current holder per say, it only cares if the current thread is the current holder. This latter case doesn't need synchronization, since it's single threaded.

hmottestad: I should make some multithreaded tests though, to verify this!

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/TempTest.java
```diff
@@ -273,33 +273,6 @@ public void testShapeWithoutTargetClassValid() throws Exception {
 
 	}
 
-	@Test(expected = NoShapesLoadedException.class)
```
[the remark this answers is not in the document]
hmottestad: Yes, but I'm not sure what to do about it right now. This was initially some tests I was playing with during the very early stages of the ShaclSail. I don't know if they some of them can be deleted or not. I'll make a new issue so I don't forget to do this in the future.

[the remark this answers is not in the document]
hmottestad: https://github.com/eclipse/rdf4j/issues/1370

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
jeenbroekstra: Fwiw I added the synchronized block just out of a vague "we syncing a private property with obtaining a lock, let's avoid a race condition" without thinking too long or hard about whether that race condition was possible. Upon reflection you're probably right that it isn't necessary.

PR #212, opened 2019-03-29
Reduce amount of memory used by size()

hmottestad:
This PR addresses GitHub issue:https://github.com/eclipse/rdf4j/issues/1369 .

* don't use distinct
* 
*


compliance/store/src/test/java/org/eclipse/rdf4j/repository/sail/memory/TempTest.java
```diff
@@ -0,0 +1,111 @@
...
+		a.begin(IsolationLevels.SNAPSHOT);
+		b.begin(IsolationLevels.SNAPSHOT);
+		a.add(GUERNICA, RDF.TYPE, PAINTING);
+		a.add(JACQUELINE, RDF.TYPE, PAINTING);
+		b.prepareUpdate(QueryLanguage.SPARQL,
+			"INSERT { ?painting a <Painting> }\n" + "WHERE { [a <Painter>] <paints> ?painting "
+				+ "OPTIONAL { ?painting a ?type  } FILTER (!bound(?type)) }",
+			NS).execute();
+		a.commit();
+		assertEquals(5, size(b, null, RDF.TYPE, PAINTING, false));
```
hmottestad: The test suite says 7 here, I'm not sure what it's testing for since the result after b.commit() is 5!

@jeenbroekstra do you know why the test suite says 7?

jeenbroekstra: Er. You introduced this test case here, as far as I can tell. Where did you get it from?

jeenbroekstra: The test suite says it _should_ be 5. However, after your change (removing the distinct filter), the _actual_ result is 7, presumably due to some duplicates not being filtered out.

jeenbroekstra: It's been a while since I looked at the details of this, but broadly speaking: the optmistic sail implementations implement transaction isolation by means of a branching model - different states of the store _during concurrent transactions_ are kept in separate branches and merged as appropriate for the isolation level when being queried. 

If two open branches contain the same statements, a query will return duplicates in its result unless explicitly filtered out. Presumably, that was why the distinct filter is in place for the size operation. Since `a` is committed but `b` is not committed yet, you get duplicate results - both the committed branch and the branch for the current state of `b` will have `(Guernica, type, Painting)` and `(Jacqueline, type, Painting)`.

jeenbroekstra:
As an alternative to removing the `DistinctIteration` (which is fraught with complications), it might be interesting to have a look at how `GroupIterator` (in queryalgebra.evaluation) implements backup syncing to disk when the size of the internal set of results becomes larger than a configurable threshold. It uses mapdb for this. Might be worth extending `DistinctIteration` with a similar mechanism.

compliance/store/src/test/java/org/eclipse/rdf4j/repository/sail/memory/TempTest.java
hmottestad: I had to introduce the test case in my own code because the test is in the test suite repo and the test is written in such a way that Intellij won't run just a single test, so was hard to debug.

In this case I feel that the test is wrong!

There is a fundamental issue with the way things are implemented at the moment. Transaction isolation causes the store to act as a BAG and not a SET. Mitigating this in the size() method by calling distinct (or similar) doesn't solve the problem that getStatements will return duplicates.

I feel that getStatements should never return duplicates. However, for performance, I think that fixing getStatements the way I have done is the wrong solution.

What needs fixing is how the changeset works. The statements in approved need to only be new statements and the statements in deprecated need to only be statements that exist in the underlying store.

jeenbroekstra: > There is a fundamental issue with the way things are implemented at the moment. Transaction isolation causes the store to act as a BAG and not a SET. Mitigating this in the size() method by calling distinct (or similar) doesn't solve the problem that getStatements will return duplicates.

That doesn't make the test wrong though - the test simply encodes that we expect the `size()` operation to return the correct number (5). 

We made a design decision in the past to allow `getStatements` to work under bag semantics, so that we are not forced to do very expensive duplicate filtering all the time when only in relatively rare circumstances (concurrent transactions) duplicates will actually occur. With this decision in mind a `Distinct` was necessary to make `size()` behave conform spec.

Now if we can find a way to fix the sail source / changeset functionality so that we can guarantee no duplicates will occur in `getStatements` under any circumstances, then I'm all for it of course :)

hmottestad: Test says 7. Because of Bag. I changed it to 5.

jeenbroekstra: Ok, from which test did you copy this? Because I thought had copied/adapted this from `org.eclipse.rdf4j.repository.optimistic.MonotonicTest` (specifically `test_mergeOptionalFilterQuery`), and that test has 5 as the expectation. If there's another test out there that has 7, that might be a bug.

hmottestad: https://github.com/eclipse/rdf4j-testsuite/blob/508df49a5458bcb822a2782e740f34df3a1a1c10/store/src/main/java/org/eclipse/rdf4j/repository/optimistic/MonotonicTest.java#L573

jeenbroekstra: Ok. That is _very_ strange. I'll need to take some time to take a good look at what is being tested here. Get back to you.

jeenbroekstra: What I'm lost on is why your fix to `SailSourceConnection.size` would cause this test to start failing in the first place. In fact I just copy-pasted your fix locally and ran those tests, and it all succeeded.

jeenbroekstra: Right. So - I think you're correct that these tests are weird. The expectation clearly _should_ be 5, not 7. I haven't dug into the history but at a guess I'd say that the test was originally specified as expecting 5, then we implemented bag semantics and someone (it might even have been me) just adjusted the tests to fit actual behavior.

I'm happy for you to adjust these tests on the master branch. Please keep your changes as minimal as possible to make things green - I'd like this fix to be available for a patch release, but at the same time I'm doing some massive restructuring of tests on the develop branch, so I'd rather not have a whole heap of changes to merge from master to develop later...

hmottestad: I'll keep that in mind.

I'm not sure if this is the most efficient, or most correct, way of fixing the problem. I need to think and explore more.

hmottestad:
@rdf4j-bot run tests

jeenbroekstra:
@rdf4j-bot run tests

hmottestad:
Did you change any of the tests in the test suite that assume that getStatements is BAG instead of SET?

jeenbroekstra:
I didn't change the tests. However due to the changes I did in the overall test setup (and Jenkins as well) I thought it was a good idea to re-run.

hmottestad:
Going to target this to 3.0.0 instead. Since this is technically a breaking change.

hmottestad:
# Benchmark results

## Before BAG/SET optimisation
```
Benchmark                                              (isolationLevel)  Mode  Cnt   Score   Error  Units
MemoryBenchmark.simpleFoafModel                                    NONE  avgt   10   4.213 ± 0.085  ms/op
MemoryBenchmark.simpleFoafModel                        READ_UNCOMMITTED  avgt   10   9.451 ± 0.070  ms/op
MemoryBenchmark.simpleFoafModel                          READ_COMMITTED  avgt   10   8.963 ± 0.124  ms/op
MemoryBenchmark.simpleFoafModel                           SNAPSHOT_READ  avgt   10   9.263 ± 0.250  ms/op
MemoryBenchmark.simpleFoafModel                                SNAPSHOT  avgt   10   9.666 ± 0.372  ms/op
MemoryBenchmark.simpleFoafModel                            SERIALIZABLE  avgt   10   9.159 ± 0.147  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                           NONE  avgt   10   9.513 ± 0.174  ms/op
MemoryBenchmark.simpleFoafModelDuplicate               READ_UNCOMMITTED  avgt   10  20.227 ± 0.325  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                 READ_COMMITTED  avgt   10  20.404 ± 0.516  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                  SNAPSHOT_READ  avgt   10  22.491 ± 1.362  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                       SNAPSHOT  avgt   10  22.365 ± 1.301  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                   SERIALIZABLE  avgt   10  22.745 ± 2.431  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple                   NONE  avgt   10  12.454 ± 0.446  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple       READ_UNCOMMITTED  avgt   10  32.848 ± 4.497  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple         READ_COMMITTED  avgt   10  32.748 ± 3.152  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple          SNAPSHOT_READ  avgt   10  32.099 ± 3.909  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple               SNAPSHOT  avgt   10  33.108 ± 3.844  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple           SERIALIZABLE  avgt   10  33.783 ± 3.715  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush              NONE  avgt   10  12.556 ± 0.735  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush  READ_UNCOMMITTED  avgt   10  33.651 ± 4.173  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush    READ_COMMITTED  avgt   10  32.779 ± 4.359  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush     SNAPSHOT_READ  avgt   10  32.067 ± 2.532  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush          SNAPSHOT  avgt   10  32.777 ± 3.899  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush      SERIALIZABLE  avgt   10  33.134 ± 2.611  ms/op
MemoryBenchmark.simpleFoafModelSize                                NONE  avgt   10   5.920 ± 0.327  ms/op
MemoryBenchmark.simpleFoafModelSize                    READ_UNCOMMITTED  avgt   10  11.408 ± 0.756  ms/op
MemoryBenchmark.simpleFoafModelSize                      READ_COMMITTED  avgt   10  11.166 ± 0.613  ms/op
MemoryBenchmark.simpleFoafModelSize                       SNAPSHOT_READ  avgt   10  11.349 ± 0.845  ms/op
MemoryBenchmark.simpleFoafModelSize                            SNAPSHOT  avgt   10  11.106 ± 0.745  ms/op
MemoryBenchmark.simpleFoafModelSize                        SERIALIZABLE  avgt   10  11.373 ± 0.674  ms/op
```

## After BAG/SET optimisation
```
Benchmark                                              (isolationLevel)  Mode  Cnt    Score   Error  Units
MemoryBenchmark.simpleFoafModel                                    NONE  avgt   10    4.387 ± 0.089  ms/op
MemoryBenchmark.simpleFoafModel                        READ_UNCOMMITTED  avgt   10    9.434 ± 0.242  ms/op
MemoryBenchmark.simpleFoafModel                          READ_COMMITTED  avgt   10    9.630 ± 0.269  ms/op
MemoryBenchmark.simpleFoafModel                           SNAPSHOT_READ  avgt   10    9.311 ± 0.182  ms/op
MemoryBenchmark.simpleFoafModel                                SNAPSHOT  avgt   10    9.473 ± 0.152  ms/op
MemoryBenchmark.simpleFoafModel                            SERIALIZABLE  avgt   10    9.582 ± 0.180  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                           NONE  avgt   10    9.468 ± 0.195  ms/op
MemoryBenchmark.simpleFoafModelDuplicate               READ_UNCOMMITTED  avgt   10   26.793 ± 0.655  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                 READ_COMMITTED  avgt   10   29.798 ± 0.364  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                  SNAPSHOT_READ  avgt   10   29.728 ± 1.282  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                       SNAPSHOT  avgt   10   31.210 ± 0.476  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                   SERIALIZABLE  avgt   10   31.099 ± 0.251  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple                   NONE  avgt   10   13.318 ± 0.386  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple       READ_UNCOMMITTED  avgt   10   99.127 ± 1.132  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple         READ_COMMITTED  avgt   10   99.024 ± 2.937  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple          SNAPSHOT_READ  avgt   10  102.197 ± 1.890  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple               SNAPSHOT  avgt   10   99.097 ± 2.036  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple           SERIALIZABLE  avgt   10   99.883 ± 1.494  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush              NONE  avgt   10   12.057 ± 0.855  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush  READ_UNCOMMITTED  avgt   10   83.544 ± 2.137  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush    READ_COMMITTED  avgt   10   83.968 ± 3.267  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush     SNAPSHOT_READ  avgt   10   81.468 ± 1.570  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush          SNAPSHOT  avgt   10   83.387 ± 1.201  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush      SERIALIZABLE  avgt   10   84.549 ± 1.715  ms/op
MemoryBenchmark.simpleFoafModelSize                                NONE  avgt   10    4.531 ± 0.129  ms/op
MemoryBenchmark.simpleFoafModelSize                    READ_UNCOMMITTED  avgt   10    9.745 ± 0.209  ms/op
MemoryBenchmark.simpleFoafModelSize                      READ_COMMITTED  avgt   10   10.231 ± 0.812  ms/op
MemoryBenchmark.simpleFoafModelSize                       SNAPSHOT_READ  avgt   10   10.898 ± 0.198  ms/op
MemoryBenchmark.simpleFoafModelSize                            SNAPSHOT  avgt   10   10.509 ± 0.270  ms/op
MemoryBenchmark.simpleFoafModelSize                        SERIALIZABLE  avgt   10   10.822 ± 0.244  ms/op
```

hmottestad:
@jeenbroekstra I have a new approach now that I think might work.

I have created a distinct union operator that takes the iterator from the base sail, the "added statements" model and a function to filter that model (instead of taking a filtered model, which is a bit slow). For every Statement in the base sail iterator, it gets removed from the "added statements" model. When the base sail iterator is empty, the "added statements" model gets filtered and an iterator on that supplies the final elements.

Essentially it will dedupe the "added statements" model, for statements that the user requests from the base sail. It is lazy, as in it will only dedupe if the user calls getStatements (or via SPARQL) after adding statements in the same transaction. The dedupe also persists in the "added statements" model, so it only needs to be deduped once. Subsequent calls don't need to dedupe.

For efficiency this needs to be merged: https://github.com/eclipse/rdf4j/pull/1417 It creates a no-op for remove(...) from an empty LinkedHashModel.

This code isn't clean. The benchmark names as horrible. And there might be bugs when removing and adding the same statement. What I'm wondering about is if you think this is a reasonable approach @jeenbroekstra ? It creates a small cost for all getStatements (or SPARQL) called on transactions that have added data, but this is about as minimal as I can manage.

sail-base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
```diff
@@ -279,6 +286,61 @@ public void handleClose() throws SailException {
 		}
 	}
 
+	private CloseableIteration<? extends Statement, SailException> distinctModelReducingUnion(
```
hmottestad: this is the relevant code

sail-base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
```diff
@@ -279,6 +286,61 @@ public void handleClose() throws SailException {
...
+			public Statement next() throws SailException {
+				calculateNext();
+				Statement temp = next;
+				next = null;
+				return temp;
+			}
+
+			@Override
+			public void remove() throws SailException {
+
```
hmottestad: do I need to implement remove() ?

hmottestad:
Also, so far I haven’t had any issues with the ShaclSail and using multiple threads for read operations on the same connection. This might break that.

hmottestad:
@rdf4j-bot run tests

hmottestad:
Some more cases that the iterator needs to handle.

hmottestad:
Added some locking and isolation. Here are the new benchmarks:

```
# Run complete. Total time: 00:25:30

Benchmark                                                                               (isolationLevel)  Mode  Cnt   Score   Error  Units
MemoryBenchmark.simpleFoafModel                                                                     NONE  avgt   10   3.637 ± 0.076  ms/op
MemoryBenchmark.simpleFoafModel                                                         READ_UNCOMMITTED  avgt   10   8.538 ± 0.118  ms/op
MemoryBenchmark.simpleFoafModel                                                           READ_COMMITTED  avgt   10   9.000 ± 0.227  ms/op
MemoryBenchmark.simpleFoafModel                                                            SNAPSHOT_READ  avgt   10   8.528 ± 0.088  ms/op
MemoryBenchmark.simpleFoafModel                                                                 SNAPSHOT  avgt   10   8.571 ± 0.263  ms/op
MemoryBenchmark.simpleFoafModel                                                             SERIALIZABLE  avgt   10   8.661 ± 0.195  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                                                            NONE  avgt   10   8.257 ± 0.224  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                                                READ_UNCOMMITTED  avgt   10  18.372 ± 0.364  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                                                  READ_COMMITTED  avgt   10  19.134 ± 0.760  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                                                   SNAPSHOT_READ  avgt   10  18.501 ± 0.486  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                                                        SNAPSHOT  avgt   10  18.879 ± 0.526  ms/op
MemoryBenchmark.simpleFoafModelDuplicate                                                    SERIALIZABLE  avgt   10  19.130 ± 0.431  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple                                                    NONE  avgt   10  11.583 ± 0.523  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple                                        READ_UNCOMMITTED  avgt   10  23.236 ± 2.165  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple                                          READ_COMMITTED  avgt   10  22.852 ± 0.823  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple                                           SNAPSHOT_READ  avgt   10  23.022 ± 1.241  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple                                                SNAPSHOT  avgt   10  22.250 ± 0.816  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultiple                                            SERIALIZABLE  avgt   10  22.517 ± 0.458  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush                                               NONE  avgt   10  11.149 ± 0.398  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush                                   READ_UNCOMMITTED  avgt   10  22.531 ± 0.587  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush                                     READ_COMMITTED  avgt   10  22.706 ± 0.963  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush                                      SNAPSHOT_READ  avgt   10  22.719 ± 1.001  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush                                           SNAPSHOT  avgt   10  22.241 ± 0.428  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleFlush                                       SERIALIZABLE  avgt   10  22.955 ± 0.742  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatement                                    NONE  avgt   10  12.630 ± 0.434  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatement                        READ_UNCOMMITTED  avgt   10  45.911 ± 1.810  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatement                          READ_COMMITTED  avgt   10  48.409 ± 5.709  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatement                           SNAPSHOT_READ  avgt   10  47.323 ± 1.262  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatement                                SNAPSHOT  avgt   10  45.927 ± 1.107  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatement                            SERIALIZABLE  avgt   10  42.389 ± 0.958  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatementIteratorMatchesNothing              NONE  avgt   10   8.836 ± 0.282  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatementIteratorMatchesNothing  READ_UNCOMMITTED  avgt   10  19.574 ± 0.458  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatementIteratorMatchesNothing    READ_COMMITTED  avgt   10  20.206 ± 0.751  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatementIteratorMatchesNothing     SNAPSHOT_READ  avgt   10  20.360 ± 0.537  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatementIteratorMatchesNothing          SNAPSHOT  avgt   10  20.045 ± 0.955  ms/op
MemoryBenchmark.simpleFoafModelDuplicateMultipleWithNewStatementIteratorMatchesNothing      SERIALIZABLE  avgt   10  20.463 ± 1.421  ms/op
MemoryBenchmark.simpleFoafModelSize                                                                 NONE  avgt   10   4.175 ± 0.134  ms/op
MemoryBenchmark.simpleFoafModelSize                                                     READ_UNCOMMITTED  avgt   10   8.995 ± 0.170  ms/op
MemoryBenchmark.simpleFoafModelSize                                                       READ_COMMITTED  avgt   10   9.249 ± 0.405  ms/op
MemoryBenchmark.simpleFoafModelSize                                                        SNAPSHOT_READ  avgt   10  10.153 ± 0.207  ms/op
MemoryBenchmark.simpleFoafModelSize                                                             SNAPSHOT  avgt   10   9.146 ± 0.231  ms/op
MemoryBenchmark.simpleFoafModelSize                                                         SERIALIZABLE  avgt   10   9.232 ± 0.196  ms/op
```

hmottestad:
@jeenbroekstra there are 144 test failures and I'm not sure why. All in:

org.eclipse.rdf4j.query.parser.sparql.manifest.W3CApprovedSPARQL10QueryTest.*

I'm unable to debug any of these tests in IntelliJ, I presume it's because the are generated at runtime from the turtle manifest files.

Do you have any insight into what's going on here?

jeenbroekstra:
@hmottestad sorry for the delay. I'll take a closer look. Judging from jenkins they all fail for the exact same reason (a mismatch in the variable bindings). I'll try and take a closer look.

jeenbroekstra:
@hmottestad these failures are not caused by your feature branch - the develop branch has the same failures. It's a side effect of the remove of the Graph and GraphUtil, and the switch to using Model/Models. In other words: my fault. I'll work on a fix.

jeenbroekstra:
Turned out to be a trivial error. Fix is being deployed, I'll restart jenkins on this PR as soon as it's available.

jeenbroekstra:
@rdf4j-bot run tests

hmottestad:
So I would like approval for my approach @jeenbroekstra @barthanssens .

During a transaction there are three sources: added, removed, base.

When reading from the store, the stream of selected data get retrieved as a union from “base” and “added” (minus removed I presume). The problem is that a statement can be both in “base” and “added”. To fix this I do the following:

 - a distinct union operator on “base” and “added” iterates over the iterator from “base” (lazy)
 - each statement returned to the user is removed from “added” (lazy)
 - when the iterator on “base” is empty a new iterator is opened on “added” (not lazy)

The iterators are selective and only return the data that matches the specified triple pattern. 

To handle concurrency issues, each “added.remove(...)” is synchronised. And when retrieving from “added”, the entire filtered iterator is consumed into an array list (synchronised). This stops concurrent modification exceptions.

Benefits to this approach:
 - almost everything is lazy
 - only transactions where the users adds/updates data in the transaction and then reads/updates are affected
 - transactions that only read are not affected at all
 - data is deduped once - and only once

Issues with this approach:
 - reads on big transactions could fail (no spilling to disk)
 - reads on big transactions will be slower

hmottestad:
Btw. This code needs a fair bit of cleanup. I just need some validation that this approach is fine before I continue.

memory/src/test/java/org/eclipse/rdf4j/sail/memory/DuplicateStatementTest.java
```diff
@@ -0,0 +1,132 @@
...
+import org.eclipse.rdf4j.common.iteration.Iterations;
+import org.eclipse.rdf4j.model.Statement;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.sail.NotifyingSailConnection;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.stream.Stream;
+
+public class DuplicateStatementTest {
```
jeenbroekstra: I'd like to see these same tests applied to the native store - perhaps these tests can be moved in the `RDFStoreTest` abstract test class.

sail-base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
jeenbroekstra: Could you add the explanation of what this method does (basically the things you mentioned in your github comment) as Javadoc for this method? It looks good to me, but it will probably help future developer understand why it's set up this way.

sail-base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
```diff
@@ -279,6 +286,69 @@ public void handleClose() throws SailException {
...
+	private CloseableIteration<? extends Statement, SailException> distinctModelReducingUnion(
+			CloseableIteration<? extends Statement, SailException> iter, Model originalModel,
+			Function<Model, Model> filterable) {
+
+		return new CloseableIteration<Statement, SailException>() {
+
+			Statement next;
+			Iterator<Statement> filteredStatementsIterator;
+
+			private void calculateNext() {
```
jeenbroekstra: I think this can be refactored as an extension of `LookAheadIteration`. Your `calculateNext` is basically its `getNextElement` method.

hmottestad:
I’ll try to get this sorted today. Even though I proposed the code freeze for m1 for yesterday it would still be nice to get into the 3.0 release.

jeenbroekstra:
I don't think we freeze completely until 3.0 final is out the door, just for getting the m1 ready. Would be good to get this in, but if not, no worries, there's plenty of time to get it in before the final 3.0 release.

sail-base/src/main/java/org/eclipse/rdf4j/sail/base/SailDatasetImpl.java
hmottestad: Can you point me to that code @jeenbroekstra. I found a couple of LookAheadIteration classes but none that look similar.

hmottestad: never mind, I think I found it

hmottestad: jepp, found it....really simplified the code. Thanks @jeenbroekstra . I think I might end up using this that in the ShaclSail too.

hmottestad:
@rdf4j-bot run tests

hmottestad:
this is ready to merge now once jenkins clears it

hmottestad:
@rdf4j-bot run tests

PR #214, opened 2019-03-31
Issues/1290 only revalidate modified shapes

hmottestad:

* compare node shapes so not all are run when only some are changed
* 
*


hmottestad:
#### Before:
```
Benchmark                                            Mode  Cnt    Score   Error  Units
ShaclLoadingBenchmark.testAddingDataAndAddingShapes  avgt   10  105.803 ± 2.262  ms/op
```

#### After:
```
Benchmark                                            Mode  Cnt   Score   Error  Units
ShaclLoadingBenchmark.testAddingDataAndAddingShapes  avgt   10  26.364 ± 0.615  ms/op
```

hmottestad:
Due to the way the java code is structured, changing a single property shape inside a node shape will cause the entire node shape to be revalidated. Waiting until after sh:node is implemented before optimising this.

PR #222, opened 2019-04-10
Issues/1388 validate files that do not fit in memory

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1388 .

* clean up utf-8 in all code
* benchmark with large generated file, nativestore and memory restrictions
* fix for not being able to load a single big file in a single transaction into nativestore
* fixes in SHACL to be cache-less (eg. no caching needed anywhere)

### DIFF
--- a/sail-api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
+++ b/sail-api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
@@ -7,17 +7,6 @@
  *******************************************************************************/
 package org.eclipse.rdf4j.sail.helpers;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.IdentityHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
-
 import org.eclipse.rdf4j.IsolationLevel;
 import org.eclipse.rdf4j.IsolationLevels;
 import org.eclipse.rdf4j.common.iteration.CloseableIteration;
@@ -39,9 +28,20 @@ import org.eclipse.rdf4j.sail.UpdateContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.IdentityHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
 /**
  * Abstract Class offering base functionality for SailConnection implementations.
- * 
+ *
  * @author Arjohn Kampman
  * @author Jeen Broekstra
  */
@@ -130,7 +130,7 @@ public abstract class AbstractSailConnection implements SailConnection {
 
 	/**
 	 * Verifies if a transaction is currently active. Throws a {@link SailException} if no transaction is active.
-	 * 
+	 *
 	 * @throws SailException if no transaction is active.
 	 */
 	protected void verifyIsActive() throws SailException {
@@ -149,6 +149,8 @@ public abstract class AbstractSailConnection implements SailConnection {
 		if (level == null) {
 			level = this.sailBase.getDefaultIsolationLevel();
 		}
+
+		addedCount = 0;
 
 		IsolationLevel compatibleLevel = IsolationLevels.getCompatibleIsolationLevel(level,
 				this.sailBase.getSupportedIsolationLevels());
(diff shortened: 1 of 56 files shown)


sail-api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
```diff
@@ -429,12 +432,21 @@ public final void rollback() throws SailException {
...
 			flushPendingUpdates();
 		}
+
 		addStatement(null, subj, pred, obj, contexts);
+		if (transactionIsolationLevel == IsolationLevels.NONE) {
+			addedCount++;
+			if (addedCount % 10000 == 0) {
+				flushPendingUpdates();
+			}
+		}
```
hmottestad: @jeenbroekstra Can I get your feedback on this?

I found out that NativeStore, even with IsolationLevels.NONE will cache all added statements in a TreeModel that overflows to disk. This was so slow that I gave up waiting. Flushing periodically (eg. every 10 000 statements) will write the changeset to disk and create a new one. This keeps the changeset in memory and utilises the efficient disk storage in the NativeStore instead of the inefficient overflow in the TreeModel.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/AST/OrPropertyShape.java
```diff
@@ -64,15 +65,15 @@ public PlanNode getPlan(ShaclSailConnection shaclSailConnection, NodeShape nodeS
...
-		BufferedSplitter targetNodesToValidate;
+		PlanNodeProvider targetNodesToValidate;
 		if (overrideTargetNode == null) {
 			targetNodesToValidate = new BufferedSplitter(unionAll(initialPlanNodes.stream()
 					.flatMap(Collection::stream)
 					.map(p -> new TrimTuple(p, 0, 1)) // we only want the targets
 					.collect(Collectors.toList())));
 
 		} else {
-			targetNodesToValidate = new BufferedSplitter(overrideTargetNode);
```
hmottestad: Implementing a PlanNodeProvider system allowed this BufferedSplitter to be removed, which means we don't cache anything here. Might actually want to reintroduce this and make it configurable from the cache setting in the ShaclSail

sail-api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
```diff
@@ -429,12 +432,21 @@ public final void rollback() throws SailException {
 		}
 	}
 
+	int addedCount = 0;
```
hmottestad: probably make this private and move it to the top

hmottestad:
I've tried a few things to improve performance, like pushing filters and adjusting various pieces of code. Measurably this hasn't made any difference. The biggest performance boost will probably come from either not requiring ordering in the queries, or bulking them in some way. The NativeStore shouldn't really need to sort anything though, since it uses a B-tree that is naturally sorted.

sail-api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
jeenbroekstra: I don't think this is the right place to do this. `AbstractSailConnection` is a "sensible default" implementation that multiple Sail implementations (including third party ones) rely on. Determining at what point an active transaction should flush to free up space is up to the specific sail implementation, and not something we should enforce like this at the abstract level (especially in a method that's `final` and therefore can't even be overridden). 

If there is a fix to be made to the native store's caching strategy, it should probably happen in SailSourceConnection, or in NativeStoreConnection itself. 

I should also like to know why the MemoryOverfowModel is turning out to be so inefficient. It's designed to not sync to disk until the JVM heap becomes full, in which case it should seamlessly switch to overflow to a disk store - this should be fairly efficient. I've noticed it is currently set up to just duplicate the indexing strategy of the master store, which makes little sense I think - a minimal index should be enough, we're just streaming triples to disk temporarily after all.

hmottestad: I was fairly certain that it was the wrong place yes. But it was fairly hard while debugging to find a place where I could implement this at all. 

I will look more at this. 

In the mean time I will try to produce some benchmark results to show performance with and without the flushing. 

And a quick question. Does IsolationLevels.NONE have to support rollback? Is there any way to add data in bulk and skip all the transactional support. You know, purely for performance.

hmottestad: Performance before this chunking was implemented:

```
# Warmup Iteration   1: 502120.334 ms/op
```
Thats 502 seconds or 8 minutes.

After the implementation of chunking:

```
# Warmup Iteration   1: 16522.558 ms/op
```
So 16 seconds.

This is the pure "loading into native store" benchmark without any SHACL.

jeenbroekstra: No, rollback support is not guaranteed at NONE level. But like you discovered, I think our native store supports more than is strictly necessary. Keep in mind that these isolation levels are about a _minimal_ guarantee.

That said I wouldn't be against modifications to the native store that made bulk upload in NONE level more efficient.

hmottestad: At the moment I will just move my code into the NativeStoreConnection, which seems to work fine. I don't feel confident enough right now to dig into why the current overflow is so slow or how to make IsolationLevels.NONE bypass the branching approach.

hmottestad: Is this ok for now @jeenbroekstra ?

PR #225, opened 2019-04-13
Issues/1375 bigger benchmark real file

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1375 .

* real file for benchmark
* 
*


hmottestad:
Still waiting for CQ

http://dev.eclipse.org/ipzilla/show_bug.cgi?id=19489

hmottestad:
Screenshot 2019-04-28 at 08 41 11

hmottestad:
@jeenbroekstra I got the CQ approved. What do I do now?

Does the CC0-1.0 and CC-BY-2.0 need to be incorporated into our license?

jeenbroekstra:
CC0 is essentially "public domain", so we don't really need to do anything about that. CC-BY-2.0, however, requires attribution and an indication if any changes (from the original) were made. 

Even if we don't distribute this as part of a release it's in our public github repo, so we should do _some_ attribution I think. I suggest that next to the data file we add a markdown file called `LICENSE-datagovbe.md`, and add the relevant information there.

PR #233, opened 2019-04-29
init optional in Sail and SailRepository 

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#1223 .

* updated SailRepository to conform to modify Repository API contract
* made Sail.init() optional in the same way
* added unit test for Abstract Sail and fixed various implementations


hmottestad:
This is nice. Lazy automatic init.

jeenbroekstra:
> This is nice. Lazy automatic init.
> 
> It does open up for more race conditions though. Are there tests for multiple threads calling getConnections at the same time?

@hmottestad There are testsuites for things like concurrent transaction handling of course, but to be fair all of those use explicitly initialized sails right now. So I've added a basic unit test to AbractsSail which spins up 100 concurrent threads that all try to get a connection at roughly the same time. It's always tough to simulate possible deadlock scenarios but this is a decent smoke test at least, I think.

PR #242, opened 2019-05-22
Issues/1419 serializable shacl

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1419 .

* refactoring of the ShaclSailConnection to extract code into logical classes and helpers
* double transaction with locking to validate SNAPSHOT as if it was SERIALIZABLE
* bechmarks

### DIFF
--- a/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
+++ b/shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
@@ -36,6 +36,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.StampedLock;
 
@@ -159,7 +160,8 @@ public class ShaclSail extends NotifyingSailWrapper {
 	private boolean validationEnabled = ShaclSailConfig.VALIDATION_ENABLED_DEFAULT;
 	private boolean cacheSelectNodes = ShaclSailConfig.CACHE_SELECT_NODES_DEFAULT;
 	private boolean rdfsSubClassReasoning = ShaclSailConfig.RDFS_SUB_CLASS_REASONING_DEFAULT;
-	private boolean performanceLogging = false;
+	private boolean serializableValidation = ShaclSailConfig.SERIALIZABLE_VALIDATION_DEFAULT;
+	private boolean performanceLogging = ShaclSailConfig.PERFORMANCE_LOGGING_DEFAULT;
 
 	static {
 		try {
@@ -175,10 +177,6 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 	}
 
-	private static String resourceAsString(String s) throws IOException {
-		return IOUtils.toString(ShaclSail.class.getClassLoader().getResourceAsStream(s), StandardCharsets.UTF_8);
-	}
-
 	public ShaclSail(NotifyingSail baseSail) {
 		super(baseSail);
 
@@ -186,6 +184,22 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 	public ShaclSail() {
 		super();
+	}
+
+	// This is used to keep track of the current connection, if the opening and closing of connections is done serially.
+	// If it is done in parallel, then this will catch that and the multipleConcurrentConnections == true.
+	private transient ShaclSailConnection currentConnection;
+	private transient boolean multipleConcurrentConnections;
+
+	synchronized void closeConnection(ShaclSailConnection connection) {
+		if (connection == currentConnection) {
+			currentConnection = null;
+		}
+	}
+
+	synchronized boolean usesSingleConnection() {
+//		return false; // if this method returns false, then the connection will always use the new serializable validation
+		return !multipleConcurrentConnections;
 	}
 
 	/**
@@ -255,6 +269,8 @@ public class ShaclSail extends NotifyingSailWrapper {
 			}
 			path = path + "-shapes-graph/";
 
+			logger.info("Shapes will be persisted in: " + path);
+
 			shapesRepo = new SailRepository(new MemoryStore(new File(path)));
 		} else {
 			shapesRepo = new SailRepository(new MemoryStore());
@@ -305,8 +321,19 @@ public class ShaclSail extends NotifyingSailWrapper {
 	}
 
 	@Override
-	public NotifyingSailConnection getConnection() throws SailException {
-		return new ShaclSailConnection(this, super.getConnection(), super.getConnection(), shapesRepo.getConnection());
+	public synchronized NotifyingSailConnection getConnection() throws SailException {
+
+		ShaclSailConnection shaclSailConnection = new ShaclSailConnection(this, super.getConnection(),
+				super.getConnection(), super.getConnection(), super.getConnection(),
+				shapesRepo.getConnection());
+
+		if (currentConnection == null) {
+			currentConnection = shaclSailConnection;
+		} else {
+			multipleConcurrentConnections = true;
+		}
+
+		return shaclSailConnection;
 	}
 
 	List<NodeShape> getNodeShapes() {
@@ -356,13 +383,17 @@ public class ShaclSail extends NotifyingSailWrapper {
 
 	/**
 	 * Releases the exclusive write lock.
+	 *
+	 * @return
 	 */
-	void releaseExclusiveWriteLock(long stamp) {
+	long releaseExclusiveWriteLock(long stamp) {
+		assert stamp != 0;
 		threadHoldingWriteLock = null;
 		lock.unlockWrite(stamp);
+		return 0;
 	}
 
-	long readlock() {
+	long acquireReadlock() {
 		if (threadHoldingWriteLock == Thread.currentThread()) {
 			throw new SailConflictException(
 					"Deadlock detected when a single thread uses multiple connections " +
@@ -372,8 +403,23 @@ public class ShaclSail extends NotifyingSailWrapper {
 		return lock.readLock();
 	}
 
-	void releaseReadlock(long stamp) {
+	long releaseReadlock(long stamp) {
+		assert stamp != 0;
+
 		lock.unlockRead(stamp);
+		return 0;
+	}
+
+	synchronized long convertToReadLock(long writeLockStamp) {
+		assert writeLockStamp != 0;
+		long l = lock.tryConvertToReadLock(writeLockStamp);
+		assert l != 0;
+		threadHoldingWriteLock = null;
+		return l;
+	}
+
+	void setNodeShapes(List<NodeShape> nodeShapes) {
+		this.nodeShapes = nodeShapes;
 	}
 
 	/**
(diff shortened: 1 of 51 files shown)


hmottestad:
@jeenbroekstra this branch here demonstrates an issue with validating graphs in SNAPSHOT isolation.

Essentially this is the problem:

shape: rdfs:Resouce max 2 rdfs:label (max 2 labels per rdfs:Resource)

Threads: T1, T2

- T1.begin
- T1.add
  - ex:resource1 a rdfs:Resource
  - ex:resource1 rdfs:label "a", "b".
- T2.begin
- T2.add
  - ex:resource1 a rdfs:Resource
  - ex:resource1 rdfs:label "c", "d".
- T1.commit
- T2.commit

This passes fine, because no single thread violates the shapes. It should not pass, because  ex:resource1 now has 4 labels.

I'm wondering if maybe simple locking will be sufficient to fix this. Exclusive lock starting with validation until the transaction is complete. This will naturally affect throughput, but if it affects throughput less than using SERIALIZABLE it might be good enough.

jeenbroekstra:
I'm not 100% sure either but it sounds sensible. I'm also not opposed to just "upgrade" the transaction level to SERIALIZABLE if people try to use SNAPSHOT, btw.

hmottestad:
I’ve given this a go to see what options there are. And the following approach works:

Before committing to the base sail. In a synchronized call. Start a new transaction, add and remove the same statements. Call prepare to validate SHACL. 

This works because we essentially serialize each commit() call. And peek at the state the base sail will be in when our commit goes through, so we know if changes that have happened between our .begin() and .commit() will affect our validation. 

Multithreaded benchmark against Serializable shows this to be around 3-4 times faster on my laptop. This is because Serializable will fail a transaction if another transaction has modified the base sail during its execution. So rerunning those failed transactions forces everything to be serial in practice. 

It is slower for non-multithreaded workloads though. Since it in essence runs the SHACL validation twice!

I still have a lot to figure out. 
 1. Can I make it faster for non-multithreaded workloads?
 2. How to handle changes to the shapes 
 3. Should this be enabled by default?

hmottestad:
With a bunch of refactoring I've made it so that the validation happens in prepare() instead of regular validation. So now slightly faster than using SERIALIZABLE (per transaction) as well as considerably faster for multithreaded applications.

I still need to work a lot on optimising the locking.

hmottestad:
I’m quite content with this approach now. So the next phase is some git cleanup, renaming variables and classes so that everything is consistent. Then trying to see if there are any edge cases I should make tests for.

hmottestad:
Benchmarks.

 - ParallelBenchmark.shaclSerializable uses connection.begin(SERIALIZABLE).
 - ParallelBenchmark.shaclSnapshot uses connection.begin(SNAPSHOT) together with the locking and double transactions to mimick a serializable transaction during validation
 - ParallelBenchmark.shaclSnapshotWithoutSerializableValidation is just the plain old connection.begin(SNAPSHOT) without anything else fancy

```
Benchmark                                                     Mode  Cnt    Score    Error  Units
ParallelBenchmark.shaclSerializable                           avgt   10  202.374 ± 20.122  ms/op
ParallelBenchmark.shaclSnapshot                               avgt   10   54.061 ±  2.380  ms/op
ParallelBenchmark.shaclSnapshotWithoutSerializableValidation  avgt   10   25.641 ±  2.327  ms/op
```

hmottestad:
@rdf4j-bot run tests

hmottestad:
Todo
 - native store benchmark

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/BackwardsChainingConnection.java
```diff
@@ -0,0 +1,128 @@
...
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.sail.SailConnection;
+import org.eclipse.rdf4j.sail.SailException;
+import org.eclipse.rdf4j.sail.helpers.SailConnectionWrapper;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class BackwardsChainingConnection extends SailConnectionWrapper {
```
hmottestad: add some docs to this, and would be nice if we could manage to make it protected

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/VerySimpleBackwardsChainingConnection.java
```diff
@@ -0,0 +1,119 @@
...
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.sail.SailConnection;
+import org.eclipse.rdf4j.sail.SailException;
+import org.eclipse.rdf4j.sail.helpers.SailConnectionWrapper;
+
+import java.util.HashSet;
+import java.util.Set;
+
+class VerySimpleBackwardsChainingConnection extends SailConnectionWrapper {
```
hmottestad: add some docs to this

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/planNodes/AbstractBulkJoinPlanNode.java
```diff
@@ -45,59 +44,60 @@ void runQuery(ArrayDeque<Tuple> left, ArrayDeque<Tuple> right, SailConnection co
...
-
-
-			})
-			.map(tuple -> tuple.line.get(0))
-			.map(v -> (Resource) v)
-			.map(r -> new ListBindingSet(Collections.singletonList("a"), Collections.singletonList(r)))
-			.collect(Collectors.toList());
-	}
+				.filter(tuple -> {
+					System.out.println("HERE");
```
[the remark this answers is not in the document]
hmottestad: ahh, yes

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSailConnection.java
```diff
@@ -113,8 +116,11 @@ public void begin(IsolationLevel level) throws SailException {
 		// successively
 		synchronized (sail) {
 			super.begin(level);
+			hasStatement(null, null, null, false); // actually force a transaction to start
```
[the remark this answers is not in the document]
hmottestad: The ShaclSail expects that the two transactions ("this", previousStateConnection)  are started against the same snapshot so that the "previousStateConnection" reflects how "this" was before it was modified. 

It might not actually matter, but I'm unable to wrap my head around what kind of effects having the two transactions start at different snapshots would have. 

I think I just forgot the fundamentals of transactions.

hmottestad:
Benchmarks: 

```
Benchmark                                                                Mode  Cnt     Score     Error  Units
ParallelBenchmark.nativeStoreShaclSerializable                           avgt   10  4380.012 ± 270.268  ms/op
ParallelBenchmark.nativeStoreShaclSnapshot                               avgt   10  1063.218 ± 186.921  ms/op
ParallelBenchmark.nativeStoreShaclSnapshotWithoutSerializableValidation  avgt   10   689.940 ± 137.278  ms/op
ParallelBenchmark.shaclSerializable                                      avgt   10   154.391 ±  18.490  ms/op
ParallelBenchmark.shaclSnapshot                                          avgt   10    57.897 ±   6.432  ms/op
ParallelBenchmark.shaclSnapshotWithoutSerializableValidation             avgt   10    26.510 ±   3.236  ms/op

```

hmottestad:
The new serializable validation adds a fair cost. However it's still considerably faster than using SERIALIZABLE as isolation level. This is due to SERIALIZABLE cancelling so many transactions.

hmottestad:
This code is now ready for review @jeenbroekstra @barthanssens .

The general approach is to start a new transaction (T2) inside a locked section (in prepare()), add/remove the statements from the current transaction (T1). Run validation. Then rollback the new transaction (T2) and commit the current one (T1).

This limits the overall throughput. Transactions can run in parallel, but run serially (one after the other) once they hit prepare() until the end of commit(). This time is mostly spent on validation. Any code the user executes (eg. SPARQL queries) between begin() and prepare/commit() can run in parallel with other connections.

The approach requires SNAPSHOT isolation and is only executed if this is the isolation level for the transaction. 

There is some code to detect if there are two connections open at the same time. This is used for optimizing if the serializable validation should be used or not. If there are no parallel transaction, the regular validation is used. 

TPS (transactions per second) on my laptop takes a  1.5x to 2.2x hit with the new serializable validation. This is still 2.7x to 4x faster than using the SERIALIZABLE isolation level.

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -564,7 +610,17 @@ public void setPerformanceLogging(boolean performanceLogging) {
 		this.performanceLogging = performanceLogging;
 	}
 
-	public void setNodeShapes(List<NodeShape> nodeShapes) {
-		this.nodeShapes = nodeShapes;
+	public boolean isSerializableValidation() {
```
hmottestad: needs docs

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/ShaclSail.java
```diff
@@ -564,7 +610,17 @@ public void setPerformanceLogging(boolean performanceLogging) {
...
+		return serializableValidation;
+	}
+
+	public void setSerializableValidation(boolean serializableValidation) {
```
hmottestad: needs docs

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/VerySimpleRdfsBackwardsChainingConnection.java
```diff
@@ -0,0 +1,124 @@
...
+import org.eclipse.rdf4j.model.vocabulary.RDF;
+import org.eclipse.rdf4j.sail.SailConnection;
+import org.eclipse.rdf4j.sail.SailException;
+import org.eclipse.rdf4j.sail.helpers.SailConnectionWrapper;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Very simple RDFS backwardschaining connection that supports type inference hasStatement and getStatement. It does not
```
hmottestad: ....type inference **on** hasState.....

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/config/ShaclSailConfig.java
```diff
@@ -185,6 +197,8 @@ public void parse(Model m, Resource implNode) throws SailConfigException {
 					.ifPresent(l -> setRdfsSubClassReasoning(l.booleanValue()));
 			Models.objectLiteral(m.filter(implNode, PERFORMANCE_LOGGING, null))
 					.ifPresent(l -> setPerformanceLogging(l.booleanValue()));
+			Models.objectLiteral(m.filter(implNode, SERIALIZABLE_VALIDATION, null))
```
hmottestad: do we have tests for this?

hmottestad:
@rdf4j-bot run tests

hmottestad:
- [ ] Also run compliance tests manually

PR #244, opened 2019-05-22
Upgrade japicmp-maven-plugin

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1421 .

* upgraded  japicmp-maven-plugin
* 
*

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -699,7 +699,7 @@
 				<plugin>
 					<groupId>com.github.siom79.japicmp</groupId>
 					<artifactId>japicmp-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


hmottestad:
@jeenbroekstra I upgrading this because the old version does not work with java 11, however the new one does. Does it require a CQ?

barthanssens:
mvn clean install -DskipTests works on JDK11, still an issue when tests are enabled

hmottestad:
@barthanssens should we just merge this anyway. As a start!

barthanssens:
+1, we need to test it anyway :-)

PR #245, opened 2019-05-24
eclipse/rdf4j#1425 only flush on removal if adds are pending

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#1425 .

* `removeStatements` needs only flush if there are pending added statements
* introduced default impl `pendingRemoved` for `AbstractConnection`
* this is WIP as some tests for SHACL fail.

The above fix brings removal operation performance from 20 seconds down to milliseconds.

### DIFF
--- a/sail-api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
+++ b/sail-api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
@@ -103,6 +103,10 @@ public abstract class AbstractSailConnection implements SailConnection {
 
 	private IsolationLevel transactionIsolationLevel;
 
+	private boolean pendingAdds;
+
+	private boolean pendingRemovals;
+
 	/*--------------*
 	 * Constructors *
 	 *--------------*/
@@ -434,6 +438,7 @@ public abstract class AbstractSailConnection implements SailConnection {
 		if (pendingRemovals()) {
 			flushPendingUpdates();
 		}
+		pendingAdds = true;
 		addStatement(null, subj, pred, obj, contexts);
 	}
 
@@ -442,6 +447,7 @@ public abstract class AbstractSailConnection implements SailConnection {
 		if (pendingAdds()) {
 			flushPendingUpdates();
 		}
+		pendingRemovals = true;
 		removeStatement(null, subj, pred, obj, contexts);
 	}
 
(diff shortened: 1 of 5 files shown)


hmottestad:
@jeenbroekstra it looks like you have test failures in the TrackAddedStatementsTest in SHACL. This actually isn't a SHACL test per say, it just tests the semantics of the NotifyingSail. I found some issues originally with how the notifying sail worked, it would in certain cases not work as expected. If I remember correctly it had to do with both adding and removing statements in the same transaction. 

At the time I introduced the method pendingRemovals() in the SailConnection interface and in the SailSourceConnection to mitigate performance issues caused by trying to uphold the semantics of the NotifyingSail.

My understanding of the semantics of the NotifyingSail:
 - it has two methods, one for added statements and one for removed
 - these are invoked continuously throughout the transaction
 - if a statements is added, then removed and then added again the methods should be invoked as follows:
   - statementAdded(st1), statementRemoved(st1), statementAdded(st1)
 - if the user attempted to remove a statement that didn't exist there should be no call to the statementRemoved method. 
  - if the user attempted to add a statement that already exists there should be no call to the statementAdded method.

shacl/src/test/java/org/eclipse/rdf4j/sail/shacl/TrackAddedStatementsTest.java
```diff
@@ -150,14 +153,15 @@ public void testValidationFailedCausesRollback() throws Exception {
 
 			try {
 				connection.commit();
-			} catch (Throwable e) {
-				System.out.println(e.getMessage());
+				fail("commit should have failed");
+			} catch (SailException e) {
```
jeenbroekstra: This is a mistake - it should be a RepositoryException. I keep getting confused by the mixed usage of Repository and Sail.

jeenbroekstra:
From what I can tell there is difference between how the memory store and the nativestore process this, which causes some test failures. I will need to take closer look to re-familiarize myself with how SailSinks and Branches are supposed to work. I _suspect_ that our implementations of `pendingRemovals` and `pendingAdds` are too tightly coupled to implementation details of the native store.

hmottestad:
It’s well worth it for performance if you can figure it out.

sail-api/src/main/java/org/eclipse/rdf4j/sail/helpers/AbstractSailConnection.java
```diff
@@ -666,6 +674,15 @@ public final void clearNamespaces() throws SailException {
 		}
 	}
 
+	@Override
+	public boolean pendingRemovals() {
+		return pendingRemovals;
+	}
+
+	protected boolean pendingAdds() {
```
hmottestad: pendingRemovals is from an interface, should pendingAdds be the same or should pendingRemovals be removed from the interface. Or maybe removed in 3.0?

jeenbroekstra: I'm thinking it should not be public. Good point. I'll mark it deprecated.

PR #253, opened 2019-06-13
eclipse/rdf4j#1453 native store context cache

jeenbroekstra:
includes persistent storage and seemless upgrade of existing stores that
miss this file.

This PR addresses GitHub issue: eclipse/rdf4j#1453 .

* added ContextStore, an in-memory index of context identifiers with on-disk persistence
* added seemless upgrade on initialization of existing native stores
* added a few unit tests

I did a few quick manual performance tests (inspired by the reported problems in eclipse/rdf4j#1452). Environment: native store with approx 10M statements, running on local linux laptop. 

Before: getContextIDs took approx. 20-30 seconds.
After: one-time initialization of the cache file took 10-20 seconds. getContextIDs took 1-2 seconds.

### DIFF
--- a/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
+++ b/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
@@ -109,7 +109,7 @@ public class ContextStore implements Iterable<Resource> {
 	 * @param context the context identifier.
 	 * @param amount  the number by which to decrease the size
 	 */
-	public void decrementBy(Resource context, int amount) {
+	public void decrementBy(Resource context, long amount) {
 		contextInfoMap.computeIfPresent(context, (c, size) -> size <= amount ? null : size - amount);
 		contentsChanged = true;
 	}
--- a/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
+++ b/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
@@ -73,12 +73,12 @@ public class ContextStore implements Iterable<Resource> {
 
 	private final ValueFactory valueFactory;
 
-	public ContextStore(NativeSailStore store, File dataDir, ValueFactory valueFactory) throws IOException {
+	public ContextStore(NativeSailStore store, File dataDir) throws IOException {
+		Objects.requireNonNull(store);
 		Objects.requireNonNull(dataDir);
-		Objects.requireNonNull(valueFactory);
 
 		this.file = new File(dataDir, FILE_NAME);
-		this.valueFactory = valueFactory;
+		this.valueFactory = store.getValueFactory();
 
 		contextInfoMap = new HashMap<>(16);
 
--- a/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
+++ b/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
@@ -195,6 +195,8 @@ public class ContextStore implements Iterable<Resource> {
 						contextInfoMap.put(context, contextSize);
 					} catch (EOFException e) {
 						break;
+					} catch (IllegalArgumentException e) {
+						throw new IOException("unable to parse context identifier: ", e);
 					}
 				}
 
--- a/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
+++ b/nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
@@ -110,7 +110,7 @@ public class ContextStore implements Iterable<Resource> {
 	 * @param amount  the number by which to decrease the size
 	 */
 	public void decrementBy(Resource context, int amount) {
-		contextInfoMap.computeIfPresent(context, (c, size) -> size == 1L ? null : size - amount);
+		contextInfoMap.computeIfPresent(context, (c, size) -> size <= amount ? null : size - amount);
 		contentsChanged = true;
 	}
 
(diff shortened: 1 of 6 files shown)


jeenbroekstra:
There's some fundamental issues with the approach, unfortunately: removing of statements may use a wildcard for the context identifier, in which case the count for specific contexts gets skewed. Getting a per-context count from the triplestore may prove difficult. 

I wonder if the simpler approach is, instead of trying to make the count sync up, just to allow empty contexts. That would make this a major new feature rather than a bug fix, but it's possible to do. If we go that route, we need to introduce an explicit method to not just clear, but also _remove_ a context. We'd also need to tweak the processing of SPARQL's `CLEAR` vs `DROP` updates.

hmottestad:
In SPARQL an empty context is the same as a non-existent context. 

Could you solve the problem with a cleaner thread or something similar?

jeenbroekstra:
Not necessarily. Sparql is uncommitted on whether empty named graphs exist or not. Which is why there is both a CLEAR and a DROP update operation in sparql. In rdf4j they do the same thing, but that's not required by the recommendation.

jeenbroekstra:
Having said that I'm open to other ways to solve this. A cleaner thread is not really possible I think since we need immediate consistency within the transaction. But I'll take a closer look at how the count is produced.

jeenbroekstra:
Found a workaround that allows a fix as a bug patch rather than a breaking change. Can I get a review please?

nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
```diff
@@ -0,0 +1,186 @@
...
+		if (contentsChanged) {
+			// Flush the changes to disk
+			writeContextsToFile();
+			contentsChanged = false;
+		}
+	}
+
+	private void writeContextsToFile() throws IOException {
+		synchronized (file) {
+			try (DataOutputStream out = new DataOutputStream(new FileOutputStream(file))) {
```
hmottestad: Should this be buffered? And what happens if you crash the application in the middle of this? Will the entire native store become corrupt?

nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
```diff
@@ -0,0 +1,186 @@
...
+		Objects.requireNonNull(store);
+		Objects.requireNonNull(dataDir);
+
+		this.file = new File(dataDir, FILE_NAME);
+		this.valueFactory = store.getValueFactory();
+
+		contextInfoMap = new HashMap<>(16);
+
+		if (file.exists()) {
+			readContextsFromFile();
```
hmottestad: Maybe this should have a try catch around it, to catch any Throwable! And rebuild if there is any kind of error?

hmottestad:
Any chance we can override the size(Resouce context) method so that it uses the cache if run in an empty transaction?

nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
jeenbroekstra: It's possible the file becomes corrupted when a crash happens mide-write, but there is a straightforward manual recovery: just delete the file. The NativeStore will automatically regenerate it. We could in future make it more sophisticated, though. 

As for buffering: I could add it but we're talking really small amounts of data here (one string and one long per context identifier in the store), so I don't think it will make a huge difference.

nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
jeenbroekstra: That's a good idea, thanks! It will make crash recovery a lot smoother.

jeenbroekstra:
> Any chance we can override the size(Resouce context) method so that it uses the cache if run in an empty transaction?

That should be possible, yes. I'll take a quick look but I may create a followup issue for that, depending on how tricky it is.

nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
hmottestad: True. Usually very little data. Would still add buffering, since it will for small data reduce IO to a single operation. One thing of note though in general, is there a chance for stale data here? Ie. there is a crash just prior to the write starts, or just after the write is successful. So that the store starts back up again with incorrect data in the context cache.

I think the only way to handle this kind of scenario is to do a flushed write to a status file to indicate that a file write on the other files has started (context cache, btree, and other files if there are any). Then when those writes are done the status file is deleted (or overwritten).

hmottestad: How about writing a map.size() as the first data line. So that the consistency of the file can be verified?

nativerdf/src/main/java/org/eclipse/rdf4j/sail/nativerdf/ContextStore.java
```diff
@@ -0,0 +1,208 @@
...
+		Objects.requireNonNull(dataDir);
+
+		this.file = new File(dataDir, FILE_NAME);
+		this.valueFactory = store.getValueFactory();
+
+		contextInfoMap = new HashMap<>(16);
+
+		try {
+			readContextsFromFile();
+		} catch (IOException e) {
```
hmottestad: What exception would be thrown if the file was corrupt? If one of the readLong or readBoolean methods failed, or if the utf8 read just read “http” and the subsequent IRI parsing failed.

jeenbroekstra: Good call. The read methods failing would all result in some form of IOException, so that's covered. However iri creation could fail with an `IllegalArgumentException`. I'll catch and wrap that.

PR #269, opened 2019-07-31
Issues/1461 shacl parallel

hmottestad:
This PR addresses GitHub issue: eclipse/rdf4j#1461 .

* enabled parallel validation as default
* add tests
*


hmottestad:
@jeenbroekstra, @barthanssens, I haven’t experienced any more problems with running validation in parallel so I think it’s time to enable it as the default. I’ve made a few more multithreaded tests and found one more bug in how prepare was being called. This last bug I reckon is deeper than just the SHACL sail, since it occurs depending on the ordering of prepare() on two separate connections/transactions.

PR #271, opened 2019-08-06
Public API clarifications for SHACL, native store and other storage modules

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#619 .


shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/SourceConstraintComponent.java
```diff
@@ -8,9 +8,16 @@
...
+import org.eclipse.rdf4j.common.annotation.InternalUseOnly;
 import org.eclipse.rdf4j.model.IRI;
 import org.eclipse.rdf4j.model.vocabulary.SHACL;
 
+/**
+ * @deprecated since 3.0. This feature is for internal use only: its existence, signature or behavior may change without
+ *             warning from one release to the next.
+ */
+@Deprecated
+@InternalUseOnly
```
hmottestad: I thought this could become stable soon.

hmottestad:
The package-info.java files, are those the java 9 module files?

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/SourceConstraintComponent.java
jeenbroekstra: Is this enum something that the user is expected to access/use directly? Or is it there to facilitate the SHACLSail's internal workings only? 

If the former, we can remove these annotations. If the latter, we should either annotate, or make the enum package-protected (if possible).

jeenbroekstra:
> The package-info.java files, are those the java 9 module files?

No, `package-info.java` is a much older mechanism, it's purely for providing metadata about packages: package-level javadoc (the first line in javadoc comment is what shows up as the one-line description on the package list in javadoc), as well as package-wide annotations. For example, you can use it to deprecate an entire package in one go (without having to mark each individual class as deprecated). 

It is a replacement for `package.html` basically - it does the same (javadoc) and more (annotations). 

See https://www.intertech.com/Blog/whats-package-info-java-for/ .

jeenbroekstra:
Build fails because `@InternalUseOnly` annotation can not be found - this is part of PR eclipse/rdf4j#1492 .

shacl/src/main/java/org/eclipse/rdf4j/sail/shacl/SourceConstraintComponent.java
hmottestad: It’s part of the “java” aspect of the validation exceptions. A validation failure creates an exception with a report in it. This report is a java object with a method to create an rdf4j model from it. For those that don’t want to deal with the model, they can use the java object directly. The type of shape that cause the failure is represented by this enum.

jeenbroekstra:
@rdf4j-bot run tests

PR #274, opened 2019-08-13
fix + regression test for eclipse/rdf4j#1498

jeenbroekstra:
This PR addresses GitHub issue: eclipse/rdf4j#1498 .

* removed cycle in init sequence
* added regression test


hmottestad:
I admit I'm unable to understand much from reading the git diff here on github, but that doesn't mean your changes aren't perfectly fine :)

PR #279, opened 2019-08-19
Added ES/Solr dependencies to javadoc maven plugin

barthanssens:
This PR addresses GitHub issue: eclipse/rdf4j#1148 .

* Set source version to 8
* Added ES and Solr dependencies to javadoc maven plugin in parent pom.xml
* Moved ES and Solr versions to parent pom.xml

Even though 'maven clean install -DskipTests' works, running 'maven clean install -DskipTests -Passembly' fails in rdf4j-storage/storage (in attach-javadocs), with 90+ errors about missing packages and symbols.

Actually this is already an issue in JDK8, and can be seen when debugging/logging, but javadoc 8 just continues after the error, so the maven goal succeeds.
But with JDK9 (and 11), it's a fatal error (even with -Xdoclint=none)

Not the prettiest patch, but I couldn't get it to work otherwise (e.g. moving this to solr/pom.xml and elasticsearch/pom.xml didn't seem to work)

Also note that the javadoc maven plugin 2.8.1 uses misspelled "additionnalDependencies" with extra "n", instead of "additionalDependencies" (fixed in newer versions)

CQs:
- [x] CQ 20696 ElasticSearch (server) 6.5.0
- [x] CQ 20697 ElasticSearch X-Content 6.5.0
- [x] CQ 20698  Solr Core 7.5.0

### DIFF
--- a/elasticsearch/pom.xml
+++ b/elasticsearch/pom.xml
@@ -15,10 +15,6 @@
 	<name>RDF4J Elastic Search Sail Index</name>
 	<description>StackableSail implementation offering full-text search on literals, based on Elastic Search.</description>
 
-	<properties>
-		<elasticsearch.version>6.5.4</elasticsearch.version>
-	</properties>
-
 	<dependencies>
 		<dependency>
 			<groupId>${project.groupId}</groupId>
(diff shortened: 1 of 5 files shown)


elasticsearch/pom.xml
```diff
@@ -43,4 +39,32 @@
...
+							<version>${elasticsearch.version}</version>
+						</additionnalDependency>
+						<additionnalDependency>
+							<groupId>org.elasticsearch</groupId>
+							<artifactId>elasticsearch-x-content</artifactId>
+							<version>${elasticsearch.version}</version>
+						</additionnalDependency>
+						<additionnalDependency>
+							<groupId>org.elasticsearch</groupId>
+							<artifactId>elasticsearch</artifactId>
```
jeenbroekstra: You said 'elasticsearch' twice :)

hmottestad:
Could we bump the javadoc version at the same time to handle that typo?

barthanssens:
Indeed, good catch.
We don't have a CQ for that, the ES-X-Content is a dependency of elasticsearch (core), which in turn is a dependency of the optional elasticsearch transport... So this probably requires a "works-with" CQ

(IIRC, it's not the doclinter but the compiler that causes the issue, since the javadoc plugin runs the compile a second time. Might be mitigated by moving from 2.8.1 to the latest javadoc version 3.1.1 that does include more options for not running compile again, but that might require more changes)

elasticsearch/pom.xml
barthanssens: Whoops :-)

barthanssens:
> Could we bump the javadoc version at the same time to handle that typo?

We can upgrade to 2.10.4 (no CQ required, already covered in the range 2.8.1-3.1.0 "works-with" dependency part of CQ20318)

I'd suggest to do that in a separate branch, since the versions of the other projects should be kept in sync (and I suppose there is no impact on the jenkins builds)

3.1.x is also an option, but probably requires a bit more changes and testing...

jeenbroekstra:
Given that this has pending CQs it will not be included in the 3.0.0 release.

jeenbroekstra:
I have tagged the 3.0 release, so we can merge this to develop now. @barthanssens are you happy for me to do so? I want to get this merged to (eventually) master (as part of the develop sync) so that it can be included in the mono-repo migration.

barthanssens:
> Given that this has pending CQs it will not be included in the 3.0.0 release.

According to IPZilla, the last CQ has just been approved, so one step close to full JDK11 support ;-)

